PR #5, opened 2026-03-24
Remove turn-taking expander and rename early results URL

tara-bogavelli:
Remove the TurnTakingAnalysis section from the Early Results page and change the URL hash from #leaderboard-oss to #early-results.


gabegma (approved):
LGTM!

PR #12, opened 2026-03-24
fix turn_strategy config

katstankiewicz:
- add turn_strategy to config so that it is loaded into PipelineConfig
- add log to confirm


gabegma (approved):
LGTM!

PR #18, opened 2026-03-26
add support for openai realtime model

katstankiewicz:
- Add vad tracking for realtime models (using external vad)
- Standardize realtime model to use api_key and model from params 
- Update .env.example to use S2S instead of realtime

### DIFF
--- a/src/eva/assistant/pipeline/services.py
+++ b/src/eva/assistant/pipeline/services.py
@@ -4,7 +4,6 @@ Creates Pipecat services with proper configuration.
 """
 
 import datetime
-import os
 from typing import Any, AsyncGenerator, Optional
 
 from deepgram import LiveOptions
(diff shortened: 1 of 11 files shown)


src/eva/assistant/pipeline/services.py
```diff
@@ -390,62 +400,54 @@ def create_realtime_llm_service(
...
+    if model_lower.startswith("openai"):
+        session_properties = get_openai_session_properties(system_prompt, params, pipecat_tools)
+        if audit_log is not None:
+            logger.info(
+                f"Using InstrumentedRealtimeLLMService for audit log interception: openai: {params.get('model')}"
+            )
+            return InstrumentedRealtimeLLMService(
+                model=params.get("model"),
+                audit_log=audit_log,
+                api_key=params.get("api_key") or os.getenv("OPENAI_API_KEY"),
```
tara-bogavelli:
```suggestion
                api_key=params["api_key"]
```
can we change this to make it consistent with the other ones. to expect api key in the params

tara-bogavelli (approved):

PR #21, opened 2026-03-26
Pr/reasoning logging

raghavm243512:
Log reasoning in agent perf stats and Audit log

### DIFF
--- a/src/eva/assistant/agentic/system.py
+++ b/src/eva/assistant/agentic/system.py
@@ -207,6 +207,7 @@ class AgenticSystem:
                     "latency": llm_stats.get("latency", 0.0),
                     "parameters": json.dumps(llm_stats.get("parameters", {})),
                     "tool_calls": json.dumps(response_tool_calls_for_stats) if response_tool_calls_for_stats else "",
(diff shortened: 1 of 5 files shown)


src/eva/assistant/agentic/system.py
```diff
@@ -207,6 +207,7 @@ async def _run_tool_loop(
                     "latency": llm_stats.get("latency", 0.0),
                     "parameters": json.dumps(llm_stats.get("parameters", {})),
                     "tool_calls": json.dumps(response_tool_calls_for_stats) if response_tool_calls_for_stats else "",
+                    "reasoning": f'"{llm_stats.get("reasoning_content", "")}"',
```
tara-bogavelli, line 210: is there are a reason we're doing the formatted string here?

tara-bogavelli (approved):

src/eva/assistant/agentic/system.py
raghavm243512: Yea, the content is put into the csv file with no delimiters around the string, so if you tried to parse the csv with something and the agent used a comma, it would break

PR #22, opened 2026-03-26
Add deepgram tts to available services

tara-bogavelli:


src/eva/assistant/pipeline/services.py
```diff
@@ -343,9 +344,18 @@ def create_tts_service(
         xtts_tts._settings.language = language_code
         return xtts_tts
 
+    elif model_lower == "deepgram":
```
gabegma: I think @katstankiewicz wanted to keep an alphabetical order in this file?

tara-bogavelli: ah ok got it. will update

gabegma (approved):
LGTM!

src/eva/assistant/pipeline/services.py
tara-bogavelli: Fixed!

PR #24, opened 2026-03-26
Improve config.json and output folder name

gabegma:
* Add model names to output folder (along with timestamp)
* Remove API keys for STT/TTS in config.json
* Reload API keys from .env on rerun


src/eva/models/config.py
```diff
@@ -269,6 +300,21 @@ class RunConfig(BaseSettings):
...
+    _PARAMS_TO_PROVIDER: ClassVar[dict[str, str]] = {
+        "stt_params": "stt",
+        "tts_params": "tts",
+        "s2s_params": "s2s",
+        "audio_llm_params": "audio_llm",
+    }
+    # Keys always read from the live environment (not persisted across runs)
+    _ENV_OVERRIDE_KEYS: ClassVar[set[str]] = {"url", "urls"}
+    # Substrings that identify secret keys (redacted in logs and config.json)
+    _SECRET_KEY_PATTERNS: ClassVar[set[str]] = {"key", "credentials", "secret"}
```
JosephMarinier, line 323: The `field_serializer`s for `model_list` and `model` only redact `key` and `credentials`, not `secret`. Have you seen an example with `secret` that we should redact? I can think of `aws_secret_access_key`, but that is already covered by `key`. That said, I guess it doesn't hurt to add `secret` to try to future-proof our implementation. Up to you!
```suggestion
    _SECRET_KEY_PATTERNS: ClassVar[set[str]] = {"key", "credentials"}
```
In any case, I think we should reuse `_SECRET_KEY_PATTERNS` or `_safe_params()` in the `field_serializer`s. What do you think?

gabegma: Good call! See my last commit.

src/eva/models/config.py
```diff
@@ -454,8 +496,12 @@ def _check_companion_services(self) -> "RunConfig":
             self._validate_service_params("TTS", self.model.tts, self.model.tts_params)
         return self
 
-    # Providers that manage their own model/key resolution (e.g. WebSocket-based)
-    _SKIP_PARAMS_VALIDATION: ClassVar[set[str]] = {"nvidia"}
+    @model_validator(mode="after")
+    def _set_default_run_id(self) -> "RunConfig":
+        if "run_id" not in self.model_fields_set:
+            suffix = "_".join(v for v in self.model.pipeline_parts.values() if v)
+            self.run_id = f"{datetime.now(UTC):%Y-%m-%d_%H-%M-%S.%f}_{suffix}"
```
JosephMarinier, line 513: I guess there may be underscores in model names/aliases? Should we have a different separator here?
Maybe double underscores?
```suggestion
            suffix = "__".join(v for v in self.model.pipeline_parts.values() if v)
            self.run_id = f"{datetime.now(UTC):%Y-%m-%d_%H-%M-%S.%f}__{suffix}"
```
Or tildes? 🤷 
```suggestion
            suffix = "~".join(v for v in self.model.pipeline_parts.values() if v)
            self.run_id = f"{datetime.now(UTC):%Y-%m-%d_%H-%M-%S.%f}~{suffix}"
```
Or is it fine as is? This is just food for thought; I don't have a strong preference.

JosephMarinier (approved):
Cool, thank you! 👍

src/eva/models/config.py
gabegma: Good point @tara-servicenow any preference here? I think I like the double underscore.

tara-bogavelli: i use - in the model names not _ (and i would like to keep this standardized for the future mode aliases if possible) so i think _ should be fine!

gabegma: Sounds good!

src/eva/models/config.py
```diff
@@ -34,9 +36,12 @@
...
 from eva.models.provenance import RunProvenance
 
+logger = logging.getLogger(__name__)
 
-def current_date_and_time():
-    return f"{datetime.now(UTC):%Y-%m-%d_%H-%M-%S.%f}"
+
+def _param_alias(params: dict[str, Any]) -> str:
+    """Return the display alias from a params dict."""
+    return params.get("alias") or params.get("model") or ""
```
tara-bogavelli:
```suggestion
    return params.get("alias") or params["model"]
```

model is required so i think you can just do this

src/eva/models/config.py
```diff
@@ -73,6 +78,15 @@ class PipelineConfig(BaseModel):
         ),
     )
 
+    @property
+    def pipeline_parts(self) -> dict[str, str]:
+        """Component names for this pipeline."""
+        return {
+            "stt": _param_alias(self.stt_params) or self.stt or "",
```
tara-bogavelli:
```suggestion
            "stt": _param_alias(self.stt_params),
```

src/eva/models/config.py
```diff
@@ -73,6 +78,15 @@ class PipelineConfig(BaseModel):
...
+            "llm": self.llm,
+            "tts": _param_alias(self.tts_params) or self.tts or "",
```
tara-bogavelli:
```suggestion
            "tts": _param_alias(self.tts_params),
```

src/eva/models/config.py
```diff
@@ -97,16 +111,10 @@ class SpeechToSpeechConfig(BaseModel):
...
-            "'smart' uses LocalSmartTurnAnalyzerV3 + SileroVAD (default). "
-            "'external' uses ExternalUserTurnStrategies for services with built-in turn detection "
-            "(e.g., deepgram-flux, Speechmatics). "
-            "Set via EVA_MODEL__TURN_STRATEGY=external."
-        ),
-    )
+    @property
+    def pipeline_parts(self) -> dict[str, str]:
+        """Component names for this pipeline."""
+        return {"s2s": _param_alias(self.s2s_params) or self.s2s}
```
tara-bogavelli, line 128:
```suggestion
        """Component names for this pipeline."""
        return {"s2s": _param_alias(self.s2s_params)}
```

src/eva/models/config.py
```diff
@@ -126,9 +134,17 @@ class AudioLLMConfig(BaseModel):
...
-    tts: str = Field(description="TTS model", examples=["cartesia", "elevenlabs"])
+    tts: str | None = Field(None, description="TTS model", examples=["cartesia", "elevenlabs"])
     tts_params: dict[str, Any] = Field({}, description="Additional TTS model parameters (JSON)")
 
+    @property
+    def pipeline_parts(self) -> dict[str, str]:
+        """Component names for this pipeline."""
+        return {
+            "audio_llm": _param_alias(self.audio_llm_params) or self.audio_llm,
+            "tts": _param_alias(self.tts_params) or self.tts or "",
```
tara-bogavelli:
```suggestion
            "tts": _param_alias(self.tts_params),
```

src/eva/models/config.py
```diff
@@ -503,20 +537,131 @@ def _expand_metrics_all(cls, v: list[str] | None) -> list[str] | None:
...
+        return data
+
+    def apply_env_overrides(self, live: "RunConfig") -> None:
+        """Apply environment-dependent values from *live* config onto this (saved) config.
+
+        Restores redacted secrets (``***``) and overrides dynamic fields (``url``,
+        ``urls``) in ``model.*_params`` and ``model_list[].litellm_params``.
+
+        Raises:
+            ValueError: If provider or alias differs for a service with redacted secrets.
```
tara-bogavelli, line 591: is there a reason we are doing both alias and provider? i can think of a scenario where provider is different but we still want to be able to use the old one (for nvidia stt for example, we have 2 ways of running that)

tara-bogavelli: i dont think we should raise an error for provider being different, i think alias should be sufficient

src/eva/models/config.py
gabegma: I believe `or self.tts` is needed - for some deployments, like kokoro, `tts_params` will just contain the url.

src/eva/models/config.py
gabegma: Ok, my concern is that there is nothing guaranteeing the alias is actually up to date with the actual model being called; it could be stale/not updated. But I could just put a warning instead of raising if we have a different provider.

src/eva/models/config.py
gabegma: Oh I guess you are saying this is no longer the case going forward? `model` is always required? What about `alias`?

tara-bogavelli: yes model is always required going forward (required in the current code here https://github.com/ServiceNow/eva/blob/main/src/eva/assistant/pipeline/services.py we do ["model"] not .get)

alias is not required. which is why i think this should be fine

params.get("alias") or params["model"]

src/eva/models/config.py
tara-bogavelli: yeah its not ideal for sure. but i think warning makes more sense.

tara-bogavelli (approved):

src/eva/models/config.py
gabegma: Done - transfer to warning for provider.

src/eva/models/config.py
gabegma: As discussed, I kept self.tts in case model and alias are both empty strings, but I removed the redundant or "", given self.tts can't be None at this stage.

src/eva/models/config.py
JosephMarinier: Oops, this change caused 13 tests to fail:
```
FAILED tests/unit/models/test_config_models.py::TestRunConfig::test_nvidia_stt_skips_params_validation - KeyError: 'model'
FAILED tests/unit/models/test_config_models.py::TestDeprecatedEnvVars::test_old_env_var_errors[env_vars-environ3-REALTIME_MODEL-EVA_MODEL__S2S-test-model-<lambda>] - KeyError: 'model'
FAILED tests/unit/models/test_config_models.py::TestDeprecatedEnvVars::test_old_env_var_errors[env_vars-environ4-EVA_MODEL__REALTIME_MODEL-EVA_MODEL__S2S-test-model-<lambda>] - KeyError: 'model'
FAILED tests/unit/models/test_config_models.py::TestDeprecatedEnvVars::test_old_env_var_errors[env_vars-environ7-REALTIME_MODEL_PARAMS-EVA_MODEL__S2S_PARAMS-value7-<lambda>] - KeyError: 'model'
FAILED tests/unit/models/test_config_models.py::TestDeprecatedEnvVars::test_old_env_var_errors[env_vars-environ8-EVA_MODEL__REALTIME_MODEL_PARAMS-EVA_MODEL__S2S_PARAMS-value8-<lambda>] - KeyError: 'model'
FAILED tests/unit/models/test_config_models.py::TestDeprecatedEnvVars::test_old_env_var_errors[env_file_vars-environ3-REALTIME_MODEL-EVA_MODEL__S2S-test-model-<lambda>] - KeyError: 'model'
FAILED tests/unit/models/test_config_models.py::TestDeprecatedEnvVars::test_old_env_var_errors[env_file_vars-environ4-EVA_MODEL__REALTIME_MODEL-EVA_MODEL__S2S-test-model-<lambda>] - KeyError: 'model'
FAILED tests/unit/models/test_config_models.py::TestDeprecatedEnvVars::test_old_env_var_errors[env_file_vars-environ7-REALTIME_MODEL_PARAMS-EVA_MODEL__S2S_PARAMS-value7-<lambda>] - KeyError: 'model'
FAILED tests/unit/models/test_config_models.py::TestDeprecatedEnvVars::test_old_env_var_errors[env_file_vars-environ8-EVA_MODEL__REALTIME_MODEL_PARAMS-EVA_MODEL__S2S_PARAMS-value8-<lambda>] - KeyError: 'model'
FAILED tests/unit/models/test_config_models.py::TestCliArgs::test_realtime_model - KeyError: 'model'
FAILED tests/unit/models/test_config_models.py::TestSpeechToSpeechConfig::test_s2s_config_from_env - KeyError: 'model'
FAILED tests/unit/models/test_config_models.py::TestSpeechToSpeechConfig::test_s2s_config_from_cli - KeyError: 'model'
FAILED tests/unit/models/test_config_models.py::TestSpeechToSpeechConfig::test_s2s_config_with_params - KeyError: 'model'
```

JosephMarinier: On the other hand, everywhere we call `_param_alias()`, we do something like
```py
_param_alias(self.stt_params) or self.stt
```
So if `params["model"]` is indeed required, the `or` cannot be reached, and we could remove it.

@tara-servicenow, @gabegma, could you advise on which change to make?

gabegma: @katstankiewicz mentioned she would fix this tomorrow :)

JosephMarinier: Ah! Awesome. Thank you!

PR #44, opened 2026-04-02
Support openai realtime framework (separate from pipecat)

tara-bogavelli:


src/eva/assistant/pipeline/nvidia_baseten.py
```diff
@@ -24,6 +24,26 @@ def _check_nvidia_available():
...
+    riva_client = None  # type: ignore[assignment]
+    _NVIDIA_AVAILABLE = False
+    NvidiaSTTService = object  # type: ignore[misc]
+    NvidiaTTSService = object  # type: ignore[misc]
+
+
+def _check_nvidia_available():
+    if not _NVIDIA_AVAILABLE:
+        raise ImportError(
+            "nvidia-riva-client is required for Baseten services. Install it with: pip install nvidia-riva-client"
```
gabegma, line 43: Why are we not adding this to our pyproject instead?

katstankiewicz: it is not compatible with pipecat 0.0.108

src/eva/assistant/gemini_live_server.py
```diff
@@ -0,0 +1,659 @@
...
+        self._running = False
+
+        # Gemini model name from s2s_params or default
+        s2s_params: dict[str, Any] = {}
+        if isinstance(self.pipeline_config, SpeechToSpeechConfig):
+            s2s_params = self.pipeline_config.s2s_params
+        else:
+            logger.error("Pipeline config is not SpeechToSpeechConfig")
+            return
+        self._model = s2s_params.get("model", "gemini-3.1-flash-live-preview")
```
tara-bogavelli: should we remove the default since i think we require model to be set in the params to make it super explicit which model you're running?

src/eva/assistant/gemini_live_server.py
```diff
@@ -0,0 +1,659 @@
...
+    # ------------------------------------------------------------------
+
+    def _create_genai_client(self) -> genai.Client:
+        """Create a google-genai Client using Vertex AI or API key."""
+        if self._api_key:
+            logger.info("Using Gemini API key for authentication")
+            return genai.Client(api_key=self._api_key)
+
+        project = os.environ.get("GOOGLE_CLOUD_PROJECT")
+        location = os.environ.get("GOOGLE_CLOUD_LOCATION", "us-central1")
```
tara-bogavelli, line 291: should we raise something if these are missing since it might not be obvious that you need these?

src/eva/assistant/gemini_live_server.py
katstankiewicz: oh yes, i agree, i will change.

src/eva/assistant/openai_realtime_server.py
```diff
@@ -0,0 +1,796 @@
...
+                        "audio": {
+                            "output": {
+                                "voice": self.pipeline_config.s2s_params.get("voice", "marin"),
+                                "format": {"type": "audio/pcm", "rate": 24000},
+                            },
+                            "input": {
+                                "format": {"type": "audio/pcm", "rate": 24000},
+                                "turn_detection": {
+                                    "type": self.pipeline_config.s2s_params.get("vad_settings", {}).get(
+                                        "type", "server_vad"
```
tara-bogavelli, line 235: love that this is all configurable now!

src/eva/assistant/openai_realtime_server.py
```diff
@@ -0,0 +1,796 @@
...
+
+        # If assistant was responding, flush interrupted response
+        if self._assistant_state.responding and self._assistant_state.transcript_parts:
+            partial_text = "".join(self._assistant_state.transcript_parts) + " [interrupted]"
+            self.audit_log.append_assistant_output(
+                partial_text,
+                timestamp_ms=self._assistant_state.first_audio_wall_ms,
+            )
+            if self._fw_log:
+                self._fw_log.tts_text(partial_text)
```
tara-bogavelli: we don't want to call this tts text anymore right?

katstankiewicz: hm we have this for both realtime and gemini-live when they are interrupted. it should contain just the part that was said so it might be helpful for debugging / noting what ended up being said

src/eva/models/config.py
```diff
@@ -331,6 +331,19 @@ class ModelDeployment(DeploymentTypedDict):
...
+        "pipecat",
+        description=(
+            "Agent framework to use for the assistant server. "
+            "'pipecat' (default): Pipecat pipeline. "
+            "'openai_realtime': OpenAI Realtime API directly. "
+            "'gemini_live': Gemini Live API via google-genai. "
+            "'elevenlabs': ElevenLabs Conversational AI. "
+            "'deepgram': Deepgram Voice Agent API."
+        ),
+    )
```
tara-bogavelli, line 427: should we remove elevenlabs and deepgram for now since we haven't supported those yet?

src/eva/orchestrator/worker.py
```diff
@@ -20,6 +20,38 @@
...
+
+        return GeminiLiveAssistantServer
+    elif framework == "elevenlabs":
+        from eva.assistant.elevenlabs_server import ElevenLabsAssistantServer
+
+        return ElevenLabsAssistantServer
+    elif framework == "deepgram":
+        from eva.assistant.deepgram_server import DeepgramAssistantServer
+
+        return DeepgramAssistantServer
```
tara-bogavelli:
```suggestion
```

.env.example
```diff
@@ -109,16 +109,16 @@ EVA_MODEL_LIST='[
 # Uncomment and add to EVA_MODEL_LIST above as needed.
 #
```
tara-bogavelli, line 111: should we add EVA_FRAMEWORK to the .env.example?

src/eva/assistant/audio_bridge.py
```diff
@@ -0,0 +1,283 @@
...
+        output_dir.mkdir(parents=True, exist_ok=True)
+
+    def write_latency(self, stage: str, value_seconds: float, model: str = "") -> None:
+        """Write a LatencyMetric entry.
+
+        Args:
+            stage: Semantic label for the stage being measured. Use ``"stt"``
+                for STT processing time, ``"tts"`` for TTS time-to-first-byte,
+                ``"model_response"`` for s2s/realtime time from user speech end
+                to first model audio chunk.
```
tara-bogavelli, line 222: i was thinking we should probably change this and make it more generalizable to write the kind of latency data that makes sense for each individual framework based on what info it offers

src/eva/assistant/openai_realtime_server.py
tara-bogavelli: @gabegma

gabegma: We are not using this in the postprocessor as of today, since we use the transcript on EL side, but I agree with saving it if we can. I think we discussed calling it something other than TTS text, though, since it's different (it's a transcript). Something like s2s_transcript maybe?

configs/prompts/simulation.yaml
```diff
@@ -125,7 +125,7 @@ audio_llm_agent:
 
 realtime_agent:
```
gabegma, line 126: Should we rename this to s2s_agent - given we are using it for Gemini live now?

src/eva/assistant/pipeline/audio_llm_processor.py
```diff
@@ -425,7 +425,7 @@ def __init__(
         base_url, _transcription_url_counter = _resolve_url(params, _transcription_url_counter)
         self._client: AsyncOpenAI = AsyncOpenAI(api_key=self._api_key, base_url=base_url)
 
-        # Callback for when transcription is ready (set by server.py)
+        # Callback for when transcription is ready (set by pipecat_server.py)
```
gabegma, line 428: Is the renaming in this file accurate? I think for some of them, it's not pipecat specific 
```suggestion
        # Callback for when transcription is ready (set by the server side)
```

src/eva/assistant/audio_bridge.py
```diff
@@ -0,0 +1,283 @@
...
+    # so that the inverse conversion recovers the original sample count.
+    expected_bytes = len(pcm_8k) * 3
+    if len(pcm_24k) < expected_bytes:
+        pcm_24k = pcm_24k + b"\x00" * (expected_bytes - len(pcm_24k))
+    elif len(pcm_24k) > expected_bytes:
+        pcm_24k = pcm_24k[:expected_bytes]
+    return pcm_24k
+
+
+def pcm16_16k_to_mulaw_8k(pcm_bytes: bytes) -> bytes:
```
gabegma: I believe this is unused

src/eva/assistant/audio_bridge.py
```diff
@@ -0,0 +1,283 @@
...
+            "event": "media",
+            "streamSid": stream_sid,
+            "media": {
+                "payload": payload,
+            },
+        }
+    )
+
+
+def create_twilio_start_response(stream_sid: str) -> str:
```
gabegma: I think this is also unused

src/eva/assistant/audio_bridge.py
```diff
@@ -0,0 +1,283 @@
...
+            },
+        }
+    )
+
+
+# ── Framework Logs Writer ────────────────────────────────────────────
+
+
+class FrameworkLogWriter:
+    """Writes framework_logs.jsonl (replacement for pipecat_logs.jsonl).
```
gabegma:
```suggestion
    """Write framework_logs.jsonl (replacement for pipecat_logs.jsonl).
```

src/eva/assistant/base_server.py
```diff
@@ -0,0 +1,244 @@
...
+            scenario_db_path=scenario_db_path,
+            tool_module_path=self.agent.tool_module_path,
+            current_date_time=self.current_date_time,
+        )
+
+        # Audio buffers for recording
+        self._audio_buffer = bytearray()
+        self.user_audio_buffer = bytearray()
+        self.assistant_audio_buffer = bytearray()
+        self._audio_sample_rate: int = 24000  # Subclasses can override
```
gabegma: I feel we have many constants for this 24000  sample rate- should we import one and use it here?

src/eva/metrics/speech_fidelity_base.py
```diff
@@ -47,6 +47,11 @@ async def compute(self, context: MetricContext) -> MetricScore:
                 )
 
             intended_turns = self._get_intended_turns(context)
+            if intended_turns and not isinstance(next(iter(intended_turns.keys())), int):
```
gabegma: Do you remember why you had this issue? I don't understand how keys cannot be integer - it would hint at a bigger bug that ideally we should fix.

katstankiewicz: i was trying to figure out why we were getting "aggregation failed" for some records so i added this to log if/when it happened. I haven't seen it get triggered so I think the aggregation failed may come from no entities being found? I couldn't figure out why for the records we got aggregation failed, but I will remove because I don't think this was an issue

src/eva/assistant/pipeline/audio_llm_processor.py
katstankiewicz: the frameprocessors are specific to pipecat

configs/prompts/simulation.yaml
katstankiewicz: I don't think this is worth changing, we use s2s and realtime interchangeably. Gemini live is a realtime model

src/eva/metrics/speech_fidelity_base.py
gabegma: Oh I see, yes this is fixed now! The code would simply put an error when normalized_score was None and wrongly assumed it was an aggregation problem.

src/eva/assistant/openai_realtime_server.py
katstankiewicz: done!

docs/assistant_server_contract.md
```diff
@@ -0,0 +1,555 @@
...
+| `self.assistant_audio_buffer` | `bytearray` | Accumulates raw assistant PCM during streaming |
+| `self._audio_buffer` | `bytearray` | Mixed audio (leave empty — base class mixes automatically) |
+| `self._audio_sample_rate` | `int` | Recording sample rate, default 24000 |
+
+After calling `super().__init__()`, narrow the config type and initialize state:
+
+```python
+def __init__(self, **kwargs):
+    super().__init__(**kwargs)
+    assert isinstance(self.pipeline_config, SpeechToSpeechConfig), (
```
gabegma: I believe we should avoid using `assert` in actual code since those can be ignored - we should probably use a raise here.

docs/assistant_server_contract.md
```diff
@@ -0,0 +1,555 @@
...
+Instantiate `FrameworkLogWriter` and `MetricsLogWriter` in `start()` and store them
+as instance attributes. Both write to files inside `self.output_dir`.
+
+### Framework log (`framework_logs.jsonl`)
+
+Record turn boundaries and assistant text for downstream processors:
+
+```python
+self._fw_log.turn_start(timestamp_ms=...)                    # when user turn begins; timestamp_ms is int | None
+self._fw_log.turn_end(was_interrupted=False, timestamp_ms=...) # when assistant finishes; timestamp_ms is int | None
```
gabegma: We are not using these turn_start/end right now, perhaps we should mention here they are optional?

docs/assistant_server_contract.md
```diff
@@ -0,0 +1,555 @@
...
+self._fw_log.llm_response(text)                              # full assistant response text
+self._fw_log.tts_text(text)                                  # text actually spoken (if interrupted, the partial text)
```
gabegma: I guess it's not exactly what was said, since there is usually more than that when there is an interruption.
```suggestion
self._fw_log.tts_text(text)                                  # text received from the TTS API (if interrupted, the partial text)
```

docs/assistant_server_contract.md
```diff
@@ -0,0 +1,555 @@
...
+self._fw_log.s2s_transcript(text, timestamp_ms=...)          # S2S transcript (what was actually spoken for s2s models that were interrupted); timestamp_ms is int | None
+```
+
+### Metrics log (`pipecat_metrics.jsonl`)
```
gabegma, line 276: Is this something we still produce? We should probably mention it's not required given we use EL?

docs/metric_context.md
```diff
@@ -314,9 +314,9 @@ Minor discrepancies are still possible (turn IDs off by one, audio timestamps no
 Benchmark Execution:
   ├─ EvaluationRecord (dataset.jsonl)
   │  ├─ user_goal, user_persona, scenario_db → MetricContext
-  │  └─ Feeds to AssistantServer + UserSimulator
+  │  └─ Feeds to PipecatAssistantServer + UserSimulator
   │
-  ├─ AssistantServer writes:
+  ├─ PipecatAssistantServer writes:
```
gabegma: I don't think this should have been renamed

src/eva/metrics/speech_fidelity_base.py
```diff
@@ -62,7 +62,7 @@ async def compute(self, context: MetricContext) -> MetricScore:
             per_turn_explanations: dict[int, str] = {}
             per_turn_transcripts: dict[int, str] = {}
             per_turn_normalized: dict[int, float] = {}
-            tts_turn_ids = sorted(intended_turns.keys())
+            tts_turn_ids = sorted(int(k) for k in intended_turns.keys())
```
gabegma:
```suggestion
            tts_turn_ids = sorted(intended_turns.keys())
```

gabegma (approved):
Colossal work @katstankiewicz and @tara-servicenow on this!! Could we bump the `simulation_version` too?

docs/assistant_server_contract.md
katstankiewicz: It's being used by _calculate_model_response_latency, but that may be out of date?

katstankiewicz: we'll keep it for now, because we have `write_latency` as well but should maybe in the future use response_latencies for `_calculate_model_response_latency`. But if someone was adding a a server that has their own stt/tts `_calculate_stt_latency` and `_calculate_tts_latency` will also look for pipecat_metrics.json

PR #45, opened 2026-04-06
Pr/configurable turn options

raghavm243512:
Few changes:

Fully configurable VAD and turn taking strategies
fix API reading bug
fix API key redaction


docs/experiment_setup.md
fanny-riols, line 1: Note: This change is from main and was already reviewed, it's not for this PR.

tara-bogavelli:
Can we add some tests for this logic? And can you also share some test runs on debug more with the different configs

raghavm243512:
@tara-servicenow 

added debug runs with turn configs under `snow.core_llm.voice_agent/tree/francis_eval/debug`
Don't think I can send full link on public comment

src/eva/models/config.py
```diff
@@ -42,7 +42,7 @@
 
 def _param_alias(params: dict[str, Any]) -> str:
     """Return the display alias from a params dict."""
-    return params.get("alias") or params["model"]
+    return params.get("alias") or params.get("model", "")
```
katstankiewicz: i think we made model required so we can revert this

src/eva/assistant/pipeline/turn_config.py
```diff
@@ -0,0 +1,123 @@
...
+    SpeechTimeoutUserTurnStopStrategy,
+    TurnAnalyzerUserTurnStopStrategy,
+)
+
+from eva.utils.logging import get_logger
+
+logger = get_logger(__name__)
+
+
+def create_vad_analyzer(vad_type: str, vad_params: dict[str, Any]) -> VADAnalyzer:
```
katstankiewicz:
```suggestion
def create_vad_analyzer(vad_type: str, vad_params: dict[str, Any]) -> VADAnalyzer | None:
```
I think we need to be able to support no vad analyzer like for flux when they handle it themselves

src/eva/assistant/pipecat_server.py
```diff
@@ -326,26 +326,42 @@ async def _realtime_tool_handler(params) -> None:
...
 
-            if (
-                isinstance(self.pipeline_config, (PipelineConfig, SpeechToSpeechConfig))
-                and self.pipeline_config.turn_strategy == "external"
-            ):
-                logger.info("Using external user turn strategies")
-                user_turn_strategies = ExternalUserTurnStrategies()
-                vad_analyzer = None
+            # Use configurable turn strategies if specified, otherwise fall back to defaults
+            if isinstance(self.pipeline_config, (PipelineConfig, AudioLLMConfig, SpeechToSpeechConfig)):
```
katstankiewicz: I think these pipelines are the only options so the else is not needed. I would remove it. Also, I think this ` turn_start_cfg = None` would fail anyway because it is not a valid turn strategy

.env.example
```diff
@@ -213,6 +213,46 @@ EVA_RECORD_IDS=
...
+
+# VAD (Voice Activity Detection) analyzer: silero
+# EVA_MODEL__VAD=silero
+
+# VAD parameters (JSON)
+# - confidence: Minimum confidence threshold (0.0-1.0, default: 0.7)
+# - start_secs: Duration to wait before confirming voice start (default: 0.2)
+# - stop_secs: Duration to wait before confirming voice stop (default: 0.2)
+# - min_volume: Minimum audio volume threshold (0.0-1.0, default: 0.6)
+# EVA_MODEL__VAD_PARAMS='{"start_secs": 0.1, "stop_secs": 0.8, "min_volume": 0.6, "confidence": 0.7}'
```
katstankiewicz:
```suggestion
# EVA_MODEL__VAD_PARAMS='{"start_secs": 0.2, "stop_secs": 0.2, "min_volume": 0.6, "confidence": 0.7}'
```
to match defaults

katstankiewicz (approved):

PR #46, opened 2026-04-07
Pr/tara/hr domain

tara-bogavelli:


scripts/run_text_only.py
```diff
@@ -493,7 +493,7 @@ async def run_record(
 
     audit_log = AuditLog()
     agent_llm_client = LiteLLMClient(model=llm_model)
-    user_sim_llm_client = LiteLLMClient(model="gpt-5.2")
+    user_sim_llm_client = LiteLLMClient(model="gpt-5.1")
```
gabegma, line 496: Was this committed on purpose?

configs/agents/medical_hr_agent.yaml
```diff
@@ -0,0 +1,1252 @@
...
+      - name: npi
+        type: string
+        description: "The provider's 10-digit NPI number. Example: 1487392045"
+      - name: case_id
+        type: string
+        description: "The case ID for the action being reported. Format: CASE-PREFIX-6digits. Example: CASE-LIC-048271"
+      - name: notification_type
+        type: string
+        enum: ["license_extension_submitted", "malpractice_updated", "privilege_reactivation"]
+        description: "The type of action being reported: license_extension_submitted, malpractice_updated, or privilege_reactivation"
```
gabegma: Detail - but I see a few description that repeats the `enum` - we could simplify

gabegma (approved):
🙇‍♀️ Colossal work - amazing Tara!!

I wonder if we should squash commits first so it doesn't make our git history on main "dirty," given there are a lot of intermediate commits?

scripts/run_text_only.py
tara-bogavelli: yes. we use gpt-5.1 as the user in voice mode, not sure why it was 5.2 here before.

configs/agents/medical_hr_agent.yaml
tara-bogavelli: Addressed!

PR #51, opened 2026-04-09
Enable more ruff rules

JosephMarinier:
Enable more Ruff rules to automatically (among other things):

- Replace `Optional[a]` with `a | None`
  - I find the name `Optional` misleading since a function argument (or a Pydantic field) may be `Optional` and "required" if it doesn't have a default value.
- Replace `from typing import` with `from collections.abc import` in cases where `typing` is deprecated
- Replace `a if a else b` with `a or b`
- Replace aliased `asyncio.TimeoutError` with `TimeoutError`
- Replace alias `Type` with `type` keyword
  - This would also cover decrecated aliases like `List` -> `list`, `Dict` -> `dict`, etc.
- Remove unnecessary mode argument `"r"` from `open()`
- Replace `f.write()` in a `for` loop with `f.writelines()`
- Replace `datetime.timezone.utc` with `datetime.UTC`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,7 +93,7 @@ target-version = "py311"
 [tool.ruff.lint]
 # https://docs.astral.sh/ruff/rules/
 # https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules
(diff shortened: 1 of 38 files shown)


gabegma (approved):
So satisfying!

PR #52, opened 2026-04-10
Results analysis app: small improvements

fanny-riols:
- Enable multiple output folders
- Toggle to select only the latest run for selected systems
- In results table, add a column for system name and one for timestamp
- Add toggle to hide failed attempts in run overview per-record table
- Split cross-run comparison table into Accuracy, EVA-X, and Diagnostic sections
- UI improvements in Record Detail

### DIFF
--- a/apps/README.md
+++ b/apps/README.md
@@ -9,13 +9,13 @@ Interactive dashboard for visualizing and comparing results.
 ### Usage
 
 ```bash
-uv run streamlit run apps/analysis.py
+streamlit run apps/analysis.py
 ```
 
 By default, the app looks for runs in the `output/` directory. You can change this in the sidebar or by setting the `EVA_OUTPUT_DIR` environment variable:
 
 ```bash
-EVA_OUTPUT_DIR=path/to/results uv run streamlit run apps/analysis.py
+EVA_OUTPUT_DIR=path/to/results streamlit run apps/analysis.py
 ```
 
 ### Views
--- a/apps/README.md
+++ b/apps/README.md
@@ -9,13 +9,13 @@ Interactive dashboard for visualizing and comparing results.
 ### Usage
 
 ```bash
-streamlit run apps/analysis.py
+uv run streamlit run apps/analysis.py
 ```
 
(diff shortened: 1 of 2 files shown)


apps/analysis.py
```diff
@@ -995,29 +1043,43 @@ def render_cross_run_comparison(run_dirs: list[Path]):
...
+    eva_x_composites = [c for c in ["EVA-X_pass", "EVA-X_mean"] if c in summary_df.columns]
+    accuracy_metrics = [m for m in ordered_metrics if _METRIC_GROUP.get(m) == "Accuracy"]
+    experience_metrics = [m for m in ordered_metrics if _METRIC_GROUP.get(m) == "Experience"]
+    other_metrics = [m for m in ordered_metrics if _METRIC_GROUP.get(m) not in {"Accuracy", "Experience"}]
+
+    id_cols = ["system_name", "run_timestamp", "run_output_dir", "records"]
+    id_rename = {
+        "system_name": "System",
+        "run_timestamp": "Timestamp",
+        "run_output_dir": "Output Dir",
```
gabegma, line 1064: Could we avoid putting the output dir in the table when we just have one? It find it adds unnecessary clutter. Multiple outputs is probably not the standard flow.

apps/README.md
```diff
@@ -9,13 +9,13 @@ Interactive dashboard for visualizing and comparing results.
 ### Usage
 
 ```bash
-streamlit run apps/analysis.py
+uv run streamlit run apps/analysis.py
```
gabegma: We didn't put `uv run` elsewhere in this README for the other commands given it's not needed if you have set your python interpreter. I would remove these changes, or apply them everywhere.

gabegma (approved):
Super!! Thank you, love the changes!

apps/analysis.py
fanny-riols: Oh yes, good point!

apps/README.md
fanny-riols: Alright

PR #54, opened 2026-04-13
Fix user turn not advancing after interruption

gabegma:
We saw an issue where user turns did not advance after an interruption, especially with S2S. This fixes that.


tara-bogavelli (approved):

katstankiewicz (approved):
lgtm!

PR #57, opened 2026-04-14
Decompose response_speed into response_speed_with_tool_calls and response_speed_no_tool_calls

fanny-riols:
## Summary

Adds a with/without tool call breakdown to the `response_speed` metric so latency can be compared between turns that required a tool call and turns that didn't.

The breakdown is exposed via a new generic `sub_metrics` field on `MetricScore` — it's not a separate registered metric and not buried in `details`. Any future metric can populate `sub_metrics` and get aggregation + UI support for free.

- `response_speed` (and `turn_taking`) both now read from a shared `context.latency_assistant_turns: dict[int, float]` (per-turn latency keyed by turn id), replacing the previous split between Pipecat's `UserBotLatencyObserver` (response_speed) and ad-hoc timestamp math (turn_taking).
- `conversation_trace` is used to classify each turn as with/without tool calls.
- `ResponseSpeedMetric` returns two entries in `sub_metrics` (`with_tool_calls`, `no_tool_calls`), each a full `MetricScore` with its own `mean_speed_seconds`, `max_speed_seconds`, `num_turns`, and `per_turn_speeds`.

Example results across 150 records:

| | Mean | Min | Max |
|---|---|---|---|
| `response_speed` (all turns) | 10.48 s | 4.57 s | 14.59 s |
| `with_tool_calls` | 11.67 s | 5.32 s | 16.37 s |
| `no_tool_calls` | 8.51 s | 3.40 s | 12.38 s |

Tool-call turns were ~3.2 s slower on average in this example.

## Generic sub-metric support

- **`MetricScore.sub_metrics: dict[str, MetricScore] | None`** — new optional field for per-metric breakdowns.
- **Runner aggregation** (`MetricsRunner._build_per_metric_aggregates`) now discovers `sub_metrics` on any metric and produces a nested `sub_metrics` entry under `per_metric.<name>` in `metrics_summary.json` (mean/min/max/count), using a shared `_aggregate_scores` helper.
- **Analysis app** flattens sub-metrics as `<metric>__<sub_key>` columns in the Diagnostic table (single-run and cross-run views). Category, color, and normalized/non-normalized flags inherit from the parent metric. A new **“Show sub-metrics”** sidebar toggle (off by default) hides these columns unless requested.

So for this PR specifically, `response_speed__with_tool_calls` and `response_speed__no_tool_calls` appear alongside `response_speed`, rendered in seconds.

## Shared per-turn latency

`turn_taking` previously computed latency inline from `audio_timestamps_user_turns` / `audio_timestamps_assistant_turns`; `response_speed` used `response_speed_latencies` (from Pipecat). Both now consume `context.latency_assistant_turns` populated by the processor, removing the duplicate computation and ensuring the two metrics agree on what “response latency” means.

## Also included

- **`apply_env_overrides`**: deployments with redacted secrets that aren't in the current `EVA_MODEL_LIST` now warn-and-skip instead of raising, as long as they aren't the active LLM. Allows metrics-only reruns in environments that don't have every deployment from the original run configured.
- **`_resolve_path`** (processor): generalized to all stored paths (audio + log files). Returns the stored path if it still exists, otherwise falls back to `output_dir/<basename>`; returns `None` when the stored path is `None` so callers can treat that as “feature disabled.” Fixes metric reruns after a run directory is moved.

image


src/eva/metrics/diagnostic/response_speed.py
```diff
@@ -63,19 +102,76 @@ async def compute(self, context: MetricContext) -> MetricScore:
...
+
+    def _get_latencies(self, context: MetricContext) -> tuple[list[float], str]:
+        return (
+            context.response_speed_latencies,
+            "No response latencies available (UserBotLatencyObserver data missing)",
+        )
+
+
+@register_metric
+class ResponseSpeedWithToolCallsMetric(_ResponseSpeedBase):
```
gabegma: Do we need these as new metrics? In my mind, they should be sub-fields computed by the response_speed metric, as we will do for turn-taking. Our number of metrics will quickly explode otherwise?

fanny-riols: Oh yes, that makes sense. Will do.

src/eva/metrics/processor.py
```diff
@@ -824,8 +837,10 @@ def _build_history(
         Each entry: {timestamp_ms, source, event_type, data}.
         """
         history = self._load_audit_log_transcript(output_dir)
-        history.extend(self._load_pipecat_logs(result.pipecat_logs_path))
-        history.extend(self._load_elevenlabs_logs(result.elevenlabs_logs_path))
+        pipecat_path = _resolve_path(result.pipecat_logs_path, output_dir / "pipecat_logs.jsonl")
```
gabegma: Could we apply this sooner rather than here? When saving the paths? to `result.pipecat_logs_path` maybe?

gabegma (approved):
LGTM!!

PR #59, opened 2026-04-15
Fix partial metric re-runs

fanny-riols:
## What

Fixes issues that made partial metric re-runs (e.g. `--metrics response_speed --force-rerun-metrics`) unreliable.

## Changes

**`metrics_summary.json` no longer overwritten on partial re-run**
When re-running a subset of metrics, `per_metric` now covers all metrics found across records rather than only the ones being re-run. `metric_errors` and `pass_at_k_config` are merged from the existing file so unrelated fields are not lost.

**LLM deployment check skipped for metrics-only re-runs**
`apply_env_overrides` gains a `strict_llm` param. Passing `strict_llm=False` when `--force-rerun-metrics` is set lets re-runs succeed on runs whose LLM deployment is no longer in `EVA_MODEL_LIST` — no simulation is needed, so the check is unnecessary.

**RunConfig loading no longer conflicts with current env pipeline mode**
`from_existing_run` used `model_validate_json` which, in pydantic-settings v2, merges env vars and `.env` on top of the saved JSON — causing a conflict if e.g. `EVA_MODEL__LLM` is set in the environment but the saved run used S2S. Fixed by loading with a local `settings_customise_sources` override that reads only from init kwargs. Also skips the pipeline mode conflict check in `_strip_other_mode_fields` when `--force-rerun-metrics` is set.


tara-bogavelli (approved):
LGTM!

PR #62, opened 2026-04-17
metric computation from DB

raghavm243512:
Marks successful auth in the DB under "session" to make consistent and usable for future domains
Auth success it considered valid if final DB["session"] value is a superset of expectation (minimum conditions)
Airline expected final DB is changed to include a session value as baseline for auth
All hashes will pop the session key before computation (so no impact to task completion or logging/saving of files)


tara-bogavelli:
One small thing - I do want authentication success to be a prerequisite for task_completion. We can modify the code for task_completion so that the requirement is now database hash equivalence and also authentication success. Otherwise this looks great!

tara-bogavelli:
One more issue here: in the airline dataset, there are a few adversarial records where the user refuses to provide the confirmation code and this is expected. In these cases, we don't authentication to happen so session in the expected scenario DB should be an empty dict. But i see in the dataset you've added auth to be expected for every single record. Please check the adversarial ones and change it to {} when we don't expect authentication to happen (for example 7.1.1 and 7.4.1 and maybe others)

src/eva/metrics/accuracy/task_completion.py
```diff
@@ -51,6 +52,21 @@ async def compute(self, context: MetricContext) -> MetricScore:
...
+        # Require auth success — if session mismatches, task cannot be complete
+        auth_mismatches = compute_session_auth_mismatches(context.expected_scenario_db, context.final_scenario_db)
+        if auth_mismatches:
+            return MetricScore(
+                name=self.name,
+                score=0.0,
+                normalized_score=0.0,
+                details={
+                    "match": False,
+                    "auth_failed": True,
```
gabegma: Could we flip the flag so it's aligned with `match`, so `auth_succeed` maybe? Could we also add `"auth_succeed": True` in the details when authentication is fine?

src/eva/metrics/diagnostic/authentication_success.py
```diff
@@ -9,58 +9,70 @@
...
 
 
+def compute_session_auth_mismatches(expected_scenario_db: dict, final_scenario_db: dict) -> dict:
+    """Check whether the final DB session satisfies the expected session.
+
+    Returns a dict of mismatched keys (empty dict means auth succeeded or no auth expected).
+    """
+    expected_session = expected_scenario_db.get("session", {})
+    if not expected_session:
+        return {}
```
gabegma: Detail, but the code below already returns an empty dict if expected_session is an empty dict. Unless `session` can be None?
```suggestion
```

src/eva/metrics/diagnostic/authentication_success.py
```diff
@@ -9,58 +9,70 @@
...
+            actual_session = context.final_scenario_db.get("session", {})
 
-            score = 1.0 if success_count > 0 else 0.0
+            if not expected_session:
+                return MetricScore(
+                    name=self.name,
+                    score=1.0,
+                    normalized_score=1.0,
+                    details={"reason": "No expected session to verify — skipping auth check"},
+                )
```
gabegma, line 65: We also have now the notion of `skipped`, so in this case I think we could put `score`/`normalized_score` to `None`, and add `skipped=True`.

tests/unit/metrics/test_authentication_success.py
```diff
@@ -13,98 +13,81 @@ def metric():
...
 
 @pytest.mark.asyncio
-async def test_no_tool_calls(metric):
-    """No tool calls at all should return 0.0 with tool not found."""
-    context = make_metric_context(tool_params=[], tool_responses=[])
-    result = await metric.compute(context)
+async def test_session_matches_expected(metric):
+    """Final session matching expected session exactly should score 1.0."""
+    ctx = make_metric_context(
+        expected_scenario_db={"session": {"confirmation_number": "ABC123", "last_name": "doe"}},
```
gabegma: Could we add this to test the .`lower()`
```suggestion
        expected_scenario_db={"session": {"confirmation_number": "ABC123", "last_name": "Doe"}},
```

gabegma (commented):
Very nice!!

I ran this on text-only mode for the airline domain and got 1 as expected everywhere.
I have small comments I think we should make (mainly changing to skipped/None).

Also, once we merge this, we won't be able to rerun metrics for old runs (since the session was not saved in the DB previously). Could this be a problem?

raghavm243512:
@gabegma yea it won't be backwards compatible, but I don't think anyone will be holding onto old runs, and that to for auth success and task completion which are least likely to fail and need reruns (no llm involved) so not something to worry about imo

gabegma (approved):
Looks great!! Thanks Raghav!

PR #64, opened 2026-04-17
Pr/rm/reasoning trace

raghavm243512:
Enables reasoning trace for anthropic, gemini, and openai models. 

Reasoning from previous turns and tool calls will be sent to subsequent turns, theoretically reducing token use (and latency) and improving accuracy. Also uses these systems in their intended way

Added reasoning tracking for these models (tracked in audit log and agent perf stats, as well as reasoning token count (in agent perf stats).


src/eva/assistant/services/llm.py
```diff
@@ -116,3 +197,247 @@ async def complete(
...
+
+                for item in response.output:
+                    if item.type == "reasoning":
+                        # Prefer human-readable summary; fall back to encrypted_content as the
+                        # reasoning-text equivalent (e.g. gpt-5.2 returns no readable summary)
+                        summary_parts = getattr(item, "summary", []) or []
+                        if summary_parts:
+                            reasoning_summary = " ".join(s.text for s in summary_parts if hasattr(s, "text"))
+                        if not reasoning_summary:
+                            reasoning_summary = getattr(item, "encrypted_content", None)
```
tara-bogavelli, line 343: **Semantic bug: don't use `encrypted_content` as human-readable reasoning text.**

When the Responses API returns no summary (the gpt-5.2 case), this falls back to the opaque `encrypted_content` blob and stores it as `reasoning_summary`. That value then flows into `stats["reasoning"]` / `stats["reasoning_content"]`, which `system.py` writes into `agent_perf_stats` and eventually the performance CSV (quoted).

The encrypted blob is a long opaque base64-ish string — the CSV's `reasoning` column will contain unreadable junk instead of reasoning text. The encrypted content is already preserved for threading via `responses_output_items`; conflating it with the human-readable reasoning field defeats the CSV's purpose.

Suggest leaving `reasoning_summary = None` when no summary exists, and letting the reasoning column legitimately be empty for gpt-5.2.

src/eva/assistant/services/llm.py
```diff
@@ -74,24 +115,64 @@ async def complete(
...
+                # For Anthropic models, extract human-readable reasoning from thinking blocks
+                if thinking_blocks and isinstance(thinking_blocks, list):
+                    # Thinking blocks are list of dicts with 'type' and 'thinking' (and optional 'signature')
+                    thinking_texts = []
+                    for block in thinking_blocks:
+                        if isinstance(block, dict) and block.get("type") == "thinking" and "thinking" in block:
+                            thinking_texts.append(block["thinking"])
+                    if thinking_texts:
+                        # Override reasoning_content with extracted thinking text
+                        reasoning_content = "\n\n".join(thinking_texts)
```
tara-bogavelli: **Redundant re-derivation.** LiteLLM already populates `reasoning_content` with the concatenated thinking text for Anthropic — the block-by-block extraction here reproduces what LiteLLM already did and then overwrites the field with the same value. The `thinking_blocks` list itself is what matters for multi-turn threading (and is captured separately below). Suggest dropping the re-extraction and relying on LiteLLM's `reasoning_content` directly.

src/eva/assistant/services/llm.py
```diff
@@ -116,3 +197,247 @@ async def complete(
...
+                    {
+                        "type": "function_call_output",
+                        "call_id": msg["tool_call_id"],
+                        "output": msg["content"],
+                    }
+                )
+            i += 1
+        return instructions, input_items
+
+    def _get_router_litellm_params(self) -> dict[str, Any]:
```
tara-bogavelli, line 262: **Inconsistent with `_lookup_use_responses_api_from_router` above.** That method catches `RuntimeError` from `router.get()`; this one doesn't, and then checks `if r is None` — which `router.get()` never returns (it raises). The None-check is dead code, and a raised RuntimeError would escape.

In practice the router is initialized when this path runs (because `use_responses_api=True` was already read from it), but pick one pattern and apply it consistently.

src/eva/assistant/services/llm.py
```diff
@@ -116,3 +197,247 @@ async def complete(
...
+                        input_items.append(
+                            {
+                                "type": "function_call_output",
+                                "call_id": tool_msg["tool_call_id"],
+                                "output": tool_msg["content"],
+                            }
+                        )
+                        j += 1
+                    i = j - 1  # will be incremented at end of loop
+                    if content := msg.get("content"):
```
tara-bogavelli: **Content placed after tool outputs.** When a historical assistant message has both `content` and `tool_calls`, the emitted order is: `function_call` → `function_call_output` → `{role: assistant, content}`. Semantically the assistant's narration came *before* tool execution, not after.

OpenAI's Responses API input is order-sensitive, and this reordering could confuse the model in multi-turn conversations that include pre-tool assistant narration. Low likelihood in practice (most turns have empty content alongside tool_calls), but worth fixing — append the content item first, then the function_call/output items.

src/eva/assistant/services/llm.py
```diff
@@ -116,3 +197,247 @@ async def complete(
...
+                    logger.warning(
+                        f"Retryable error on attempt {attempt + 1}/{max_retries + 1}: {e}. Retrying in {delay:.1f}s..."
+                    )
+                    await asyncio.sleep(delay)
+                    continue
+                else:
+                    logger.exception(f"Responses API call failed: {e}")
+                    raise
+
+        raise last_exception  # type: ignore[misc]
```
tara-bogavelli: Minor: the `# type: ignore[misc]` hints `last_exception` might be `None`, but this line is actually unreachable — non-retryable errors re-raise inside the loop, and retryable errors always set `last_exception` before the loop exits. The chat-completions branch at line 199 has the same structure without the ignore. Either both or neither.

tara-bogavelli (commented):
## Summary [Claude]

I reviewed this PR against the LiteLLM, OpenAI Responses API, Anthropic, and Gemini docs. **Overall, the implementation is well-aligned with the reference patterns** — the three tricky provider-specific behaviors (OpenAI encrypted reasoning threading, Anthropic `thinking_blocks`, Gemini `provider_specific_fields`) are each handled per docs, and pure helpers have good unit coverage.

### What's correct
- `store=False` + `include=["reasoning.encrypted_content"]` + echoing `response.output` back as input items matches the [OpenAI cookbook stateless pattern](https://developers.openai.com/cookbook/examples/responses_api/reasoning_items) exactly.
- `reasoning={"effort": ...}` parameter shape is correct for the Responses API.
- Attaching `thinking_blocks` to the assistant message in history is exactly what [LiteLLM docs prescribe](https://docs.litellm.ai/docs/reasoning_content) for Anthropic extended-thinking + tool-calling.
- Switching to `tc.model_dump(exclude_none=True)` to preserve `provider_specific_fields.thought_signature` matches LiteLLM's documented auto-preservation behavior for Gemini.
- `use_responses_api` at the top-level of the deployment rather than inside `litellm_params` is the right call — it's an EVA routing decision, not a LiteLLM param.

### Issues to address
1. **Storing encrypted reasoning blob as human-readable reasoning text** (inline comment on `llm.py`) — semantic bug that will pollute the perf CSV.
2. **Inconsistent router error handling** between `_lookup_use_responses_api_from_router` and `_get_router_litellm_params`.
3. **Historical assistant message ordering**: content placed after tool outputs in the Responses API input conversion.
4. **Redundant re-derivation** of `reasoning_content` from `thinking_blocks` (LiteLLM already populates the combined string).
5. **LiteLLM version bump** `>=1.30.0` → `>=1.82.6` — necessary for `aresponses()` and the new reasoning fields, but given recent supply-chain attacks on liteLLM, should be explicitly confirmed.
6. **Test coverage gap**: no integration test exercises `_complete_via_responses_api` end-to-end through `AgenticSystem.process_query`; the new `TestResponsesOutputItemsThreading` only mocks `llm_client.complete`. Also no evidence in the diff that a debug-mode benchmark run was performed.
7. Minor: `raise last_exception  # type: ignore[misc]` at end of `_complete_via_responses_api` papers over dead code — the chat-completions branch above has the same structure without the ignore.

See inline comments for specifics.

pyproject.toml
```diff
@@ -25,7 +25,7 @@ dependencies = [
     "elevenlabs>=1.0.0",
     "openai>=1.0.0",
     "anthropic>=0.83.0",
-    "litellm>=1.30.0",
+    "litellm>=1.82.6",
```
tara-bogavelli: LiteLLM version jump from `>=1.30.0` to `>=1.82.6` is a big step. The bump is *necessary* — `litellm.aresponses()`, `thinking_blocks`, and `provider_specific_fields` auto-preservation landed in later 1.x releases — but given recent supply-chain attacks targeting liteLLM specifically, this should be explicitly approved and pinned to a specific known-good version rather than a bare `>=`.

tests/unit/assistant/test_litellm_client.py
```diff
@@ -0,0 +1,259 @@
...
+        # Should return a message object (not a string) when there are tool calls
+        assert hasattr(result, "tool_calls")
+        assert len(result.tool_calls) == 1
+        assert result.tool_calls[0].function.name == "get_info"
+
+        # responses_output_items must be present for next-turn threading
+        output_items = stats["responses_output_items"]
+        assert output_items is not None
+        assert any(it["type"] == "reasoning" for it in output_items)
+        assert any(it["type"] == "function_call" for it in output_items)
```
tara-bogavelli, line 277: **Coverage gap:** The new tests cover the pure helpers (`_convert_tools_for_responses_api`, `_convert_messages_for_responses_api`) and reasoning extraction with mocked `litellm.aresponses`. But there's no integration test that exercises `_complete_via_responses_api` end-to-end through `AgenticSystem.process_query` — the `TestResponsesOutputItemsThreading` test mocks `llm_client.complete` directly, so it never goes through the real Responses API code path.

src/eva/assistant/services/llm.py
raghavm243512: I kept this since I think we should prove that something happened

tests/unit/assistant/test_litellm_client.py
raghavm243512: not sure a good way to do this one unless we actually call the model

tara-bogavelli:
should we be switching to the responses API? i get an error when i try to run gpt-5.4 with reasoning 

2026-04-22 22:14:07,299 | ERROR    | eva.assistant.services.llm:179 | LiteLLM completion failed: litellm.BadRequestError: OpenAIException - Function tools with reasoning_effort are not supported for gpt-5.4 in /v1/chat/completions. Please use /v1/responses instead.. Received Model Group=gpt-5.4

raghavm243512:
@tara-servicenow that error is actually why the switch is needed

completions doesn't support reasoning + tool calls

this does. You need to add `"use_responses_api": true` (see .env.example changes in this PR)

tara-bogavelli:
ahh got it makes sense sorry i missed that earlier!

src/eva/assistant/services/llm.py
```diff
@@ -23,21 +26,47 @@ class LiteLLMClient:
...
+        The field lives at the top level of the deployment object (not inside litellm_params),
+        since it is an EVA routing decision rather than a LiteLLM parameter.
+
+        Returns False if the router is not yet initialized or the deployment is not found.
+        """
+        try:
+            r = router.get()
+        except RuntimeError:
+            return False
+        for deployment in getattr(r, "model_list", []):
```
JosephMarinier: If the router has not been initialized before we get here, we have a real problem. The `router.get()` will raise a `RuntimeError`, and we want to see it. I think the ideal would be to let the error through and stop the execution.

Similarly, I would use `r.model_list` instead of `getattr(r, "model_list", [])`, because if the router no longer has a `model_list` one day, we want to catch it rather than skip silently.
```suggestion
        """
        for deployment in router.get().model_list:
```

src/eva/assistant/services/llm.py
```diff
@@ -23,21 +26,47 @@ class LiteLLMClient:
...
+            use_responses_api: Whether to route calls through the OpenAI Responses API instead of
+                chat completions. Required for multi-turn encrypted reasoning on OpenAI o-series /
+                gpt-5.x models. When None (default), the value is read from the ``use_responses_api``
+                top-level field in the matching EVA_MODEL_LIST deployment. Pass True/False explicitly
+                to override.
         """
         self.model = model
+        self.use_responses_api = (
+            self._lookup_use_responses_api_from_router() if use_responses_api is None else use_responses_api
+        )
```
JosephMarinier: The `use_responses_api` argument is only used in unit tests, right? I'm concerned about modifying the code just to make the tests work. I'd prefer changing the tests to pass it properly via the `model_list`. I'll commit something.
```suggestion
    def __init__(self, model: str):
"""Initialize LiteLLM client.

       Args:
           model: Model name matching a model_name in EVA_MODEL_LIST (e.g., 'gpt-5.2', 'gemini-3-pro')
       """
        self.model = model
        self.use_responses_api = self._lookup_use_responses_api_from_router()
```

pyproject.toml
JosephMarinier: Don't forget to lock the dependencies with `uv lock` (or `uv sync` does it automatically). I'll push it.

src/eva/assistant/services/llm.py
JosephMarinier: Done in 35c7ac1b410cb7c14655415f3afc23b672346ede

src/eva/assistant/services/llm.py
```diff
@@ -46,6 +75,9 @@ async def complete(
         Args:
             messages: List of message dicts with 'role' and 'content'
             tools: Optional list of tools in OpenAI format
+            reasoning_effort: Optional reasoning effort level ("low", "medium", "high").
+                            If provided, enables extended thinking for models that support it.
+                            Can also be set in model config via litellm_params.reasoning_effort.
```
JosephMarinier: Are we using this anywhere?

raghavm243512: I would've thought so but litellm looks like it covers this too. I'll go over the params again and just fix up anything duplicated.

JosephMarinier (approved):
Cool! Thank you!

PR #70, opened 2026-04-21
Penalize no agent response

tara-bogavelli:
Tested on run with perturbations, before 9/50 would rerun due to "inactivity timeout", with this change they were all identified as agent errors and no reruns were needed based on the conversation finished check. The diagnostic metric shows those 9 failed on agent turn response, all other succeeded. I looked at a few of the examples it flagged and listened to the audio and it was as expected, the agent did not respond to a user turn that i could clearly hear.


docs/metrics/conversation_correctly_finished.md
```diff
@@ -0,0 +1,68 @@
...
+
+Uses the following MetricContext fields:
+- `conversation_ended_reason`: Reason the conversation terminated (e.g. `"goodbye"`, `"inactivity_timeout"`, `"error"`).
+- `audio_timestamps_user_turns`: Per-turn `(start, end)` audio intervals for the user.
+- `audio_timestamps_assistant_turns`: Per-turn `(start, end)` audio intervals for the assistant.
+
+### Evaluation Methodology
+
+1. Compute `last_audio_speaker` as whichever side (`"user"` or `"assistant"`) has the latest audio end-timestamp across all turns. Returns `None` if neither side recorded audio.
+2. Flag the record as a missed turn iff `conversation_ended_reason == "inactivity_timeout"` **and** `last_audio_speaker == "user"`.
```
gabegma:
```suggestion
2. Flag the record as a missed turn if `conversation_ended_reason == "inactivity_timeout"` **and** `last_audio_speaker == "user"`.
```

src/eva/orchestrator/validation_runner.py
```diff
@@ -59,93 +79,95 @@ def __init__(
...
+
+        for record_id in check_ids:
+            ctx = contexts.get(record_id)
+            if ctx is None:
+                not_finished.append(record_id)
+                continue
+            if ctx.conversation_finished:  # type: ignore[attr-defined]
+                gate_passed.append(record_id)
+                continue
+            if is_agent_timeout_on_user_turn(
```
gabegma: I had in mind that we would modify the `conversation_valid_end` definition to include either the agent timeout _or_ that the end tool call is properly called. Any reason for doing it manually here rather than in the metric directly?

src/eva/orchestrator/runner.py
```diff
@@ -925,6 +932,14 @@ def settings_customise_sources(cls, settings_cls, init_settings, **kwargs):
                 return (init_settings,)
 
         config_data = json.loads(config_path.read_text())
+        # Backwards compat: remap any legacy metric names saved in an older config.json.
+        from eva.metrics.legacy_aliases import rename_metric_keys, rename_metric_list
```
gabegma: Could we move the import to the top? I think we are importing 3 times now.

src/eva/orchestrator/validation_runner.py
```diff
@@ -59,93 +79,95 @@ def __init__(
...
+        gate_passed: list[str] = []
+        not_finished: list[str] = []
+        agent_timeout: set[str] = set()
+
+        for record_id in check_ids:
+            ctx = contexts.get(record_id)
+            if ctx is None:
+                not_finished.append(record_id)
+                continue
+            if ctx.conversation_finished:  # type: ignore[attr-defined]
```
gabegma: Could we rename this context variable to `conversation_valid_end` since I think this is what it now represents?

gabegma (approved):
LGTM!! Minor detail but we could update the doc for conversation_valid_end now that it includes agent time-out failures.

PR #73, opened 2026-04-23
Improve airline agent: enum schemas, policy consistency, same-day fee fix

gabegma:
## Summary

- Enrich `get_reservation` responses with per-segment origin/destination/scheduled times from the journeys table (read-time only, no DB impact).
- Promote enum-valued tool params to proper JSON Schema `enum:` fields so the LLM gets a hard constraint; add a two-way drift test between the YAML and the Pydantic StrEnums.
- Require caller confirmation before invoking `transfer_to_agent`.
- Make standby policy consistent: free for all fare classes; drop the unexercised Basic Economy exclusion and the now-meaningless elite standby-fee waivers.
- Fix `same_day` vs `voluntary` rebook fees: Basic Economy now pays $199 same-day / $75 voluntary; Business/First pay $75 same-day / $0 voluntary. Clarify in instructions when to pick each.

## Test plan

- [x] `pytest tests/unit/airline/` — 110 passed
- [x] New drift test catches YAML↔Pydantic enum mismatches (verified by intentionally leaving one out)
- [x] Verified no `expected_scenario_db` impact: `change_fee` is response-only; no existing scenario touches the changed code paths (Business/First same-day, Basic Economy rebook)

### DIFF
--- a/configs/agents/airline_agent.yaml
+++ b/configs/agents/airline_agent.yaml
@@ -44,7 +44,7 @@ instructions: |
   
   Same-Day Changes:
   - Same-day confirmed change: $75 (waived for elite status Gold and above)
-  - Same-day standby: Free for all fare classes
+  - Same-day standby: Free (see Standby Rules)
   
   Fee Waivers:
   - IRROPS (cancellation, delay >2hrs, schedule change >2hrs): All fees waived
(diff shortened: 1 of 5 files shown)


tara-bogavelli (approved):
Looks good thanks for improving this

PR #74, opened 2026-04-23
Pronunciation Metric

oluwanifemibamgbose:
## Summary

- Adds a new **diagnostic** audio judge metric `pronunciation` that evaluates agent speech for phonetic quality (lexical stress, phoneme production) — scoped narrowly so it does not overlap with `agent_speech_fidelity` (entity/content accuracy).
- Uses Gemini as the audio judge, reusing the existing `SpeechFidelityBaseMetric` infrastructure (file upload fallback, retries, per-turn response parsing).
- Binary 0/1 rating per turn with structured per-dimension evidence (`bad_stress`, `bad_sound`)
- The judge receives audio + turn IDs only — no intended text is passed, so the judge must transcribe from audio alone.

### DIFF
--- a/configs/prompts/judge.yaml
+++ b/configs/prompts/judge.yaml
@@ -1147,98 +1147,84 @@ judge:
 
   pronunciation:
     user_prompt: |
-        You are evaluating the pronunciation quality of a voice agent.
-        Listen to the audio and assess each turn as a typical customer of a professional voice
-        agent would — not as a linguist.
+        You are evaluating the PHONETIC pronunciation quality of a voice agent.
+        Listen to the audio and assess each turn on acoustic-phonetic grounds only.
 
-        Your job is to evaluate HOW words are spoken, not whether the correct words were spoken.
-        Word-level accuracy is a separate concern — focus entirely on pronunciation.
+        This metric evaluates two narrow things: whether individual phonemes within words are
+        produced correctly, and whether lexical stress falls on the right syllable. It does
+        NOT evaluate the content of what was said, the choice of words, accent or regional
+        variation, or whether entities such as dates, currencies, acronyms, or confirmation
+        codes were rendered accurately. A separate fidelity metric handles content accuracy.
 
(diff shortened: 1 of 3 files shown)


configs/prompts/judge.yaml
```diff
@@ -1145,4 +1145,91 @@ judge:
...
+
+        - **1** (Acceptable or Great): The meaning is fully interpretable and the agent sounds
+          competent. This covers both:
+          • *Great* — pronunciation of every word, including sounds and syllable stress, is
+            in line with a native speaker of General American English. Very minor errors only.
+          • *Acceptable* — one or more pronunciation errors are present, but the turn still
+            meets a typical customer's expectations (e.g. a non-standard accent that doesn't
+            impair comprehension).
+
+        - **0** (Unacceptable): Pronunciation errors that have a high negative impact on
```
fanny-riols, line 1204: Not mentioning "bad sound" here?

configs/prompts/judge.yaml
```diff
@@ -1145,4 +1145,91 @@ judge:
...
+          substituted, omitted, added, or distorted — relative to the dictionary pronunciation.
+          Example: "ski-zors" instead of "si-zors" for "scissors". This is about the acoustic
+          realisation of phonemes, not about whether the right word was spoken.
+
+        ## What NOT to penalise
+        - Whether the agent said the correct words or rendered entities (dates, currencies,
+          acronyms, codes) accurately — those are fidelity concerns, not pronunciation.
+        - Rendering choices for numbers, dates, or amounts — e.g. "fifteen dollars" vs "one
+          five dollars" is a rendering-style decision, not a pronunciation error.
+        - Accent or regional variation — do not flag non-American vowels or rhoticity.
```
fanny-riols, line 1188: Why not flag non-American accents, for the agent? In the rating below, unde "Great", we have "in line with a native speaker of General American English.". So I find it confusing.

PR #77, opened 2026-04-24
Improvements to the results analysis app

fanny-riols:
## Summary

- **Cascading System filter**: replaced the flat Model filter in the cross-run comparison with a System filter that shows a human-readable label and is constrained by the selected Provider, so irrelevant systems are hidden automatically.
- **Complete-runs-only & hide-incomplete toggles**: two new sidebar toggles filter out partial runs from the comparison table and scatter plot; both are URL-bound so the view is shareable.
- **Per-sample heatmap**: a new interactive heatmap below the comparison tables shows per-record metric scores across all runs, with a metric selector, swap-axes toggle, and correct RdYlGn / RdYlGn_r colorscale based on whether lower or higher is better.
- **Pinned columns & row numbers**: the `#`, link, and System columns are now pinned in the dataframe so they stay visible when scrolling horizontally; the link column is labelled "Run".
- **URL-bound state**: the latest-run toggle and show-sub-metrics toggle are now bound to query params, so deep-linking to a specific view works end-to-end.
- **Lower-is-better metric set**: restricted the inverted colorscale to `response_speed` and `stt_wer` only (was applied too broadly).
- **Latest-run filter fix**: timestamp-only folders now derive their system name from `config.json` so that different models sharing the same folder format are treated as distinct systems and deduplicated correctly.


apps/analysis.py
```diff
@@ -79,6 +80,9 @@ def _build_metric_group_map() -> dict[str, str]:
 
 _NON_NORMALIZED_METRICS = {"response_speed"}
 
+# Metric names for which lower values are better
+_LOWER_BETTER_METRICS = {"response_speed", "stt_wer"}
+
```
gabegma: I detect automatically now from https://github.com/ServiceNow/eva/pull/65

apps/analysis.py
```diff
@@ -1111,15 +1181,20 @@ def _show_subtable(heading: str, composites: list, metrics: list) -> None:
...
-        sub_df.insert(0, "link", link_series)
+        sub_df.insert(0, "#", range(1, len(sub_df) + 1))
+        sub_df.insert(1, "link", link_series)
         sub_df = sub_df.rename(columns={**id_rename, **composite_rename, **col_rename})
         score_cols = [composite_rename[c] for c in composites] + [col_rename[m] for m in metrics]
         styled = sub_df.style.map(_color_cell, subset=score_cols)
         styled = styled.format(dict.fromkeys(score_cols, "{:.3f}"), na_rep="—")
         st.dataframe(
             styled,
             hide_index=True,
```
JosephMarinier: Is it important that the `#` column start at 1? If it can start at 0, we can simply show the index column by removing `hide_index=True`.
```suggestion
        sub_df.insert(0, "link", link_series)
        sub_df = sub_df.rename(columns={**id_rename, **composite_rename, **col_rename})
        score_cols = [composite_rename[c] for c in composites] + [col_rename[m] for m in metrics]
        styled = sub_df.style.map(_color_cell, subset=score_cols)
        styled = styled.format(dict.fromkeys(score_cols, "{:.3f}"), na_rep="—")
        st.dataframe(
            styled,
```

apps/analysis.py
```diff
@@ -1111,15 +1181,20 @@ def _show_subtable(heading: str, composites: list, metrics: list) -> None:
...
-            column_config={"link": st.column_config.LinkColumn(" ", display_text="🔍", width=40)},
+            column_config={
+                "#": st.column_config.NumberColumn("#", width=50, pinned=True),
```
JosephMarinier: If we apply [this suggestion](https://github.com/ServiceNow/eva/pull/77/changes#r3142374980), we can also remove this since the index column is pinned by default.
```suggestion
```

apps/analysis.py
```diff
@@ -1129,6 +1204,91 @@ def _show_subtable(heading: str, composites: list, metrics: list) -> None:
...
+            with ctrl_col:
+                selected_heatmap_metric = st.selectbox(
+                    "Metric",
+                    available_heatmap_metrics,
+                    format_func=_format_metric_name,
+                    key="heatmap_metric",
+                )
+            with swap_col:
+                st.markdown("<div style='padding-top:28px'></div>", unsafe_allow_html=True)
+                swap_axes = st.toggle("Swap axes", key="heatmap_swap_axes")
```
JosephMarinier: Streamlit now has a `st.space()` function that can be used instead of `st.markdown("<div style='padding-top:28px'></div>", unsafe_allow_html=True)`. Also, we can use `st.container(horizontal=True)` to that items flow naturally, instead of forcing arbitrary columns with `st.columns()`. Lastly, with a few tweaks, we can center the toggle to the selectbox:
```suggestion
            with st.container(horizontal=True, vertical_alignment="center", gap="medium"):
                selected_heatmap_metric = st.selectbox(
                    "Metric",
                    available_heatmap_metrics,
                    format_func=_format_metric_name,
                    key="heatmap_metric",
                )
                with st.container(width="content", gap=None):
                    st.space(28)  # Height of the selectbox's label.
                    swap_axes = st.toggle("Swap axes", key="heatmap_swap_axes")
```

JosephMarinier (approved):

gabegma (approved):

PR #91, opened 2026-04-28
Fix metric bugs

gabegma:
## Summary

Bug fixes hit while debugging a stuck rerun loop on a remote benchmark job.

### 1. Judge response parsing — `8e6342b`
`parse_judge_response` used `extract_and_load_json` (first match wins). For metrics like `faithfulness`, judge responses often contain incidental JSON fragments in prose (e.g. `[]` from tool-call references), so the first object wasn't the real answer.

Now iterates every JSON value, flattens lists, and returns the largest dict — preferring ones with a top-level `rating`. Adds a unit test for the multi-object case.

### 2. Pipeline robustness on incomplete trial dirs — `3aa4073`

- `_load_context` substituted `"{}"` for a missing `result.json`, producing a cryptic Pydantic error. Now raises `FileNotFoundError` like the existing scenario_db checks.
- `_run_and_pipeline`'s outer `except Exception: raise` was both hiding bugs and cancelling the whole iteration on a single corrupt artifact. Now catches `FileNotFoundError` / `json.JSONDecodeError` narrowly (record marked invalid, iteration continues); other exceptions still propagate.

### 3. `response_speed` skips on degenerate input — `d3dc008`, `9bb88e5`

Both error branches (all latencies filtered out by the `0 < v < 1000s` sanity range, or `latency_assistant_turns` empty because the agent never produced an overlapping turn) now return `score=None, skipped=True`. `response_speed` is diagnostic — these cases are already flagged by `task_completion`, `conversation_progression`, `turn_taking`.

### 4. `turn_taking` scores 0 on no-overlap — `9bb88e5`

When no turn has both user and assistant audio (agent didn't participate), this is a turn-taking failure, not missing data. Now scores `0.0` instead of erroring.

## Test plan
- [x] `pytest tests/unit/metrics/{test_metrics_utils,test_response_speed,test_turn_taking}.py`
- [x] `pytest tests/unit/orchestrator tests/unit/metrics/test_runner.py tests/integration/test_evaluation_mode.py` — 94 tests green

### DIFF
--- a/src/eva/metrics/utils.py
+++ b/src/eva/metrics/utils.py
@@ -10,7 +10,7 @@ from typing import Any
 from pydub import AudioSegment
 
 from eva.models.results import MetricScore
-from eva.utils.json_utils import extract_and_load_json
+from eva.utils.json_utils import extract_and_load_json, extract_and_load_json_iter
 from eva.utils.logging import get_logger
 
 logger = get_logger(__name__)
@@ -18,6 +18,11 @@ logger = get_logger(__name__)
 
 def parse_judge_response(response_text: str, record_id: str, metric_logger) -> dict | None:
     """Parse LLM judge response using robust JSON extraction.
+
+    Iterates over every JSON value found in the response (prose can contain
+    incidental JSON-like fragments such as `[]` from inline tool-arg references)
+    and returns the largest dict, preferring ones that carry a top-level
+    ``rating`` field — the structured judge answer is by far the biggest object.
 
     Args:
         response_text: Raw response from LLM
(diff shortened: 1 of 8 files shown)


src/eva/metrics/utils.py
```diff
@@ -27,26 +32,27 @@ def parse_judge_response(response_text: str, record_id: str, metric_logger) -> d
...
-    response = extract_and_load_json(response_text)
-
-    if response is None:
-        metric_logger.error(f"Failed to extract JSON from judge response for {record_id}")
+    candidates: list[dict] = []
+    for obj, _ in extract_and_load_json_iter(response_text):
+        if isinstance(obj, dict):
+            candidates.append(obj)
+        elif isinstance(obj, list):
+            candidates.extend(item for item in obj if isinstance(item, dict))
```
JosephMarinier, line 40: Do we actually expect lists of dicts? Or is that just extra safety? For example,
```py
parse_judge_response('Final ratings: [{"rating": 0}, {"rating": 100}]')
```
used to return `{"rating": 0}` and will now return `{"rating": 100}`.

tests/unit/metrics/test_metrics_utils.py
```diff
@@ -36,6 +36,24 @@ def test_invalid_json_returns_none(self):
...
+        text = (
+            "The assistant called check_room_availability with equipment_required: [] and "
+            "floor_code: ''. Final answer:\n"
+            '```json\n{"rating": 1, "dimensions": {"x": 1}}\n```'
+        )
+        result = parse_judge_response(text, "rec-1", MagicMock())
+        assert result == {"rating": 1, "dimensions": {"x": 1}}
+
+    def test_picks_largest_dict_when_multiple(self):
+        text = '{"a": 1} and later the real one: {"rating": 2, "dimensions": {"k": "v"}}'
```
JosephMarinier, line 49: I'm curious, is that an actual issue you noticed? Or did you only see an issue with the empty list, like in the test above? I wonder if the solution is simply to ignore lists and extract the first dict.

src/eva/metrics/utils.py
JosephMarinier: Resolved offline.

tests/unit/metrics/test_metrics_utils.py
JosephMarinier: Resolved offline.

JosephMarinier (approved):

PR #95, opened 2026-04-29
Enhance robustness for turn-taking metrics

nhhoang96:
- [x] Update turn-taking thresholds to be more realistic
    + Without tool call, sweet spot: [500ms,2000ms], hard late: >=3500ms, linear ramp-down: 2000ms-3500ms
    + With tool call, sweet spot: [500ms, 3000ms], hard late: 5000ms, linear ramp-down: 3000ms-5000ms
- [x] Zero-ing out turn-taking score if the conversation never finishes (keep the record progress information for analyses, only update normalized_score accordingly to reflect on aggregated score)
- [x] Add corresponding unit test cases reflecting both of the updated changes
- [x] Early/ Late thresholds are updated as halfway between sweet spot and hard spot accordingly

### DIFF
--- a/src/eva/metrics/experience/turn_taking.py
+++ b/src/eva/metrics/experience/turn_taking.py
@@ -458,16 +458,7 @@ class TurnTakingMetric(CodeMetric):
                 context.audio_timestamps_user_turns,
                 context.audio_timestamps_assistant_turns,
             )
-            conversation_completed = context.conversation_ended_reason != "error" and not missed_turn
-            incomplete_error = (
-                (
-                    "Agent failed to respond to last user turn — score zeroed"
-                    if missed_turn
-                    else "Agent failed to respond (conversation ended with error) — score zeroed"
-                )
-                if not conversation_completed
-                else None
-            )
+            incomplete_error = "Agent failed to respond to last user turn — score zeroed" if missed_turn else None
 
             details: dict[str, Any] = {
                 "per_turn_score": per_turn_score,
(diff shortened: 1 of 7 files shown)


tara-bogavelli:
We don't need to add a completed field to the MetricContext, what you should do is reuse this code, https://github.com/ServiceNow/eva/blob/main/src/eva/metrics/diagnostic/conversation_correctly_finished.py#L20-L30. Current implementation is not reading the right fields.

src/eva/metrics/experience/turn_taking.py
```diff
@@ -472,12 +473,24 @@ async def compute(self, context: MetricContext) -> MetricScore:
...
+            if not context.completed:
+                return MetricScore(
+                    name=self.name,
+                    score=0.0,
+                    normalized_score=0.0,
+                    details=details,
+                    sub_metrics=self._build_flat_sub_metrics(
+                        context, turn_keys, turns_with_tool_calls, per_turn_evidence
+                    ),
+                    error="Conversation not completed — score zeroed",
```
tara-bogavelli: Can we call this "Agent failed to respond" or "Agent failed to respond to last user turn"  instead of "Conversation not completed" to make it more clear what the error is?

tara-bogavelli:
Thanks for working on this!

src/eva/metrics/experience/turn_taking.py
nhhoang96: Great idea. I have pushed the changes!

tara-bogavelli (approved):
LGTM! Let's wait for @gabegma to review before merging though

nhhoang96 (commented):
LGTM. Thanks for fixing these!

gabegma (approved):
Thank you @nhhoang96 !

PR #96, opened 2026-04-29
Run all metrics by default

JosephMarinier:
I included a few small things before the default metrics. Let me know if anything is worth discussing in a separate PR.

* [Lock dependencies](https://github.com/ServiceNow/eva/pull/96/commits/fd65e8a655ed9f509daf35686c66e659891e1bf2) after 5f4bb76 in #93
  - [Lock dependencies in pre-commit](https://github.com/ServiceNow/eva/pull/96/commits/2d35a89e8e42bca0d80b3cfa5ff4b99b493e2515), which is also run in the CI, so `uv.lock` never gets out of sync again.
  - [Get package version dynamically from `src/eva/__init__.py`](https://github.com/ServiceNow/eva/pull/96/commits/940936e1abe1617725270a9f6dde558f702ddf64)
  - [Remove simulation and metrics versions from pyproject.toml](https://github.com/ServiceNow/eva/pull/96/commits/0b194f99822f3fc2e2f679731430f53c3fe6f884), as they are redundant with `src/eva/__init__.py` and out of sync since 5f4bb76.
* [Revert duplicated `_redact_api_keys()`](https://github.com/ServiceNow/eva/pull/96/commits/747d71a399f45034052db05fe8964350bcea2510) from #45, duplicated with the existing `_redact_model_params()`
* [Remove remaining unreachable `or`s](https://github.com/ServiceNow/eva/pull/96/commits/74832b91ce368f96f3b6bd490e7803165aa37090) left after [this](https://github.com/ServiceNow/eva/commit/6d771db15259996e20ae06bcd69274aca24d7a91#diff-5b03da5a28117137f315375ad13e2fe88f52779180b5af968805e4a24f90615bL85-R87) in #50
* [Fully remove deprecated env vars](https://github.com/ServiceNow/eva/pull/96/commits/d06f7661d9eb904e2790e5da5eab110a0c874b1d)
  - These env vars were not supported; they have been raising a deprecation error for the last 2 months, so we can assume people have stopped using them by now.
* [Update Pydantic](https://github.com/ServiceNow/eva/pull/96/commits/df958c9fdc5f34bdd57fcf6fc4a5987fb2195b2f) (this would fail without the previous commit)
* [Remove CLI aliases](https://github.com/ServiceNow/eva/pull/96/commits/20c459a22f860c3fb053b8e9a271227777199e02)
  - This is a breaking change, but specifying these old CLI aliases/shortcuts will now raise a clear `eva: error: unrecognized arguments: ...`, so this shouldn't cause silent errors.
  - [Update README](https://github.com/ServiceNow/eva/pull/96/commits/a9774f1d17fe02d3901eb6f1d207a5a040f7d801)
* [Run all metrics by default](https://github.com/ServiceNow/eva/pull/96/commits/1590ce626baf3c6b2694a4a82466e1cc8665c595)
* [Support `--metric=` (work around a bug in Pydantic)](https://github.com/ServiceNow/eva/pull/96/commits/c8eb0d4b4e5f3b67213e7045d1f2717647a643c9)


uv.lock
```diff
@@ -732,7 +732,7 @@ wheels = [
 
 [[package]]
 name = "eva"
-version = "0.1.3"
+version = "0.1.4"
```
katstankiewicz: is there a way to keep this version in sync with `src/eva/__init__.py`
this version is in sync right now but the `metrics_version` are out of sync (I mean in pyproject.toml)

katstankiewicz (approved):

uv.lock
JosephMarinier: Oh yeah! ~I can add a pre-commit to check that `src/eva/__init__.py` and `pyproject.toml` are aligned. Then, my [Lock dependencies in pre-commit](https://github.com/ServiceNow/eva/pull/96/commits/2d35a89e8e42bca0d80b3cfa5ff4b99b493e2515) should guarantee that uv.lock is aligned with `pyproject.toml`.~

JosephMarinier: Oh! Better yet: we can tell `pyproject.toml` to read the version dynamically from `src/eva/__init__.py`. 🤯 

[Get package version dynamically from src/eva/__init__.py](https://github.com/ServiceNow/eva/pull/96/commits/940936e1abe1617725270a9f6dde558f702ddf64)

JosephMarinier: I just noticed that `metrics_version` was out of sync between `pyproject.toml` and `src/eva/__init__.py`, so I did [Remove simulation and metrics versions from pyproject.toml](https://github.com/ServiceNow/eva/pull/96/commits/0b194f99822f3fc2e2f679731430f53c3fe6f884). The `src/eva/utils/provenance.py` references `src/eva/__init__.py`, and there doesn't seem to be anything referencing `pyproject.toml`. Does it make sense to remove it? That way, it can't happen again. :ok_hand: FYI @gabegma.

gabegma: yeah better! it was annoying to update in 2 places

PR #106, opened 2026-04-30
add gemini ALM support

raghavm243512:
Gemini TTS service updated to match new pipecat expectation
2nd service for ALM refactored with parent ALM class
Small bug in run_text_only fixed

Run gemini ALM with:
```
EVA_MODEL__AUDIO_LLM=gemini
EVA_MODEL__AUDIO_LLM_PARAMS='{"api_key": "NOT NEEDED but here for pydantic validation", "project":"project_name","location":"global","model":"gemini-3-flash-preview"}'
GOOGLE_APPLICATION_CREDENTIALS=path
```

### DIFF
--- a/src/eva/assistant/pipeline/services.py
+++ b/src/eva/assistant/pipeline/services.py
@@ -44,6 +44,7 @@ from pipecat.transcriptions.language import Language
 from pipecat.utils.text.base_text_filter import BaseTextFilter
 from websockets.asyncio.client import connect as websocket_connect
 
+from eva.assistant.pipeline.alm_gemini import ALMGeminiClient
 from eva.assistant.pipeline.alm_vllm import ALMvLLMClient
 from eva.assistant.pipeline.nvidia_baseten import BasetenSTTService, BasetenTTSService
 from eva.assistant.pipeline.realtime_llm import InstrumentedRealtimeLLMService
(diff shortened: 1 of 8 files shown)


src/eva/assistant/pipeline/services.py
```diff
@@ -566,27 +570,53 @@ def get_openai_session_properties(system_prompt: str, params: dict, pipecat_tool
...
+        Configured audio-LLM client.
     """
     model_lower = model.lower()
 
     # Resolve URL once (supports round-robin via "base_urls" list)
     global _audio_llm_url_counter
     base_url, _audio_llm_url_counter = _resolve_url(params, _audio_llm_url_counter)
 
+    if "gemini" in model_lower:
+        gemini_model = params.get("model", "gemini-3-flash-preview")
```
tara-bogavelli:
```suggestion
    if "gemini" in model_lower:
        gemini_model = params["model"]
```

we don't set defaults for models anymore

src/eva/assistant/pipeline/services.py
```diff
@@ -566,27 +570,53 @@ def get_openai_session_properties(system_prompt: str, params: dict, pipecat_tool
...
+            raise ValueError(
+                "Gemini audio-LLM requires either api_key (Developer API) "
+                "or project+location (Vertex AI via GOOGLE_APPLICATION_CREDENTIALS)."
+            )
+        client = ALMGeminiClient(
+            api_key=api_key,
+            base_url=base_url or None,
+            model=gemini_model,
+            temperature=params.get("temperature", 0.0),
+            max_tokens=params.get("max_tokens", 1024),
```
tara-bogavelli: is there a reason we are setting these defaults?

tara-bogavelli: they strongly recommend leaving the default of 1.0 for gemini 3 models

tara-bogavelli: shouldn't we also allow reasoning effort to be set?

raghavm243512: added reasoning effort, I have defaults just because the other alm also has them

tara-bogavelli (approved):

PR #107, opened 2026-05-01
Elevenlabs not saving

katstankiewicz:
- changes to manage elevenlabs segfault
- add assets to Dockerfile
- remove unneeded PYTHONPATH
- save result.json in worked instead of outside of semaphore


src/eva/orchestrator/worker.py
```diff
@@ -179,90 +179,128 @@ async def run(self) -> ConversationResult:
...
+                    conversation_ended_reason="error",
+                )
+
+            ended_at = datetime.now()
+
+            # Compute scenario database hashes (REQUIRED for deterministic metrics)
+            initial_db_path = self.output_dir / "initial_scenario_db.json"
+            final_db_path = self.output_dir / "final_scenario_db.json"
+
+            if not initial_db_path.exists():
```
raghavm243512: Do we need these explicit checks? open() will already raise the same exception

src/eva/assistant/elevenlabs_server.py
```diff
@@ -189,7 +189,11 @@ async def websocket_root(websocket: WebSocket):
         logger.info(f"Elevenlabs server started on ws://localhost:{self.port}")
 
     async def _shutdown(self) -> None:
-        """Stop the server, save outputs."""
+        """Stop the server (framework-specific teardown).
+
+        Note: save_outputs() is called by the base class stop() after _shutdown()
```
raghavm243512: not sure if this comment is needed (the 'Note:...'). It makes sense looking at the commit (where we see the line being removed) but not as much in the code itself

src/eva/orchestrator/worker.py
katstankiewicz: hm, yeah it will raise but not have as much detail. i'm ok with it either way. This isn't new code, just refactored, but I can still remove it to clean up a bit

raghavm243512 (approved):

PR #109, opened 2026-05-07
Refactor model config

JosephMarinier:
* [Yield multiple validation errors at once](https://github.com/ServiceNow/eva/pull/109/commits/53148a86613b653205016c7a5490265784374b57) from `_check_companion_services()`, not just the first one.
  For example, you could see
  ```
  2 validation errors for RunConfig
  model.pipeline.stt_params
    "api_key" required in EVA_MODEL__STT_PARAMS for deepgram STT
  model.pipeline.tts_params
    "model" required in EVA_MODEL__TTS_PARAMS for cartesia TTS
  ```
* [Refactor model config](https://github.com/ServiceNow/eva/pull/109/commits/6886ad60224a18ee5db6805d5c64b66feb35895d) 
  - Get rid of `_strip_other_mode_fields()` by merging the 3 model classes into a single `ModelConfig` that always accepts extra `EVA_MODEL__*_PARAMS` environment variables, even if not used (so no need to strip anything).
  - Skip validating `EVA_MODEL__*` environment variables when `EVA_MAX_RERUN_ATTEMPTS=0`. That way, if you specify `EVA_MAX_RERUN_ATTEMPTS=0`, you no longer need to remove these `EVA_MODEL__*` environment variables from your `.env` file. Somewhat-breaking change: we used to skip validating `EVA_MODEL__*` environment variables when `EVA_FORCE_RERUN_METRICS=True`, but that could be set while there are still failed conversations to rerun, so now you have to specify `EVA_MAX_RERUN_ATTEMPTS=0`.
  - This removed 82 lines of code in `src/eva/models/config.py`.

### DIFF
--- a/src/eva/metrics/runner.py
+++ b/src/eva/metrics/runner.py
@@ -149,7 +149,7 @@ class MetricsRunner:
 
         config_data = json.loads(config_path.read_text())
 
-        # Determine pipeline type from config (fallback to False for legacy runs)
+        # Determine pipeline type from config
         model_data = config_data.get("model", {})
         self._pipeline_type = get_pipeline_type(model_data) if model_data else PipelineType.CASCADE
 
--- a/src/eva/metrics/runner.py
+++ b/src/eva/metrics/runner.py
@@ -239,7 +239,7 @@ class MetricsRunner:
         for record_id, record_dir in self._discover_record_dirs(self.run_dir, self.record_ids):
             result_path = record_dir / "result.json"
             if not result_path.exists():
-                logger.info(f"process_records: {record_id} has no result.json, skipping")
+                logger.debug(f"process_records: {record_id} has no result.json, skipping")
                 continue
             try:
                 result_data = json.loads(result_path.read_text())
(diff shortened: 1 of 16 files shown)


gabegma:
Awesome! I tried it on an existing run where I had some errors, and I had an annoying error, unrelated to your changes. II will push it here - you can revert it if you want. :)

src/eva/metrics/runner.py
```diff
@@ -331,6 +331,12 @@ async def run_and_save_record(self, record_id: str, record_dir: Path) -> RecordM
         - Normal mode: only computes metrics not yet present on disk.
         - Rerun mode: only recomputes metrics that failed; never reruns already-succeeded metrics.
         """
+        # If the conversation worker never produced result.json, the record cannot
+        # be evaluated — skip cleanly instead of letting _load_context raise.
+        if not (record_dir / "result.json").exists():
+            logger.info(f"run_and_save_record: {record_id} has no result.json, skipping")
+            return None
```
JosephMarinier, line 338: @gabegma Perfect, thank you!
> I tried it on an existing run where I had some errors, and I had an annoying error, unrelated to your changes.

Did everything else work as you expected?

raghavm243512 (approved):
Looks great. 

Definitely something I should have just done when I added the ALM instead of making this mess

src/eva/metrics/runner.py
gabegma: Yes it did, sorry!! It was great :)

PR #112, opened 2026-05-12
Various bug fixes and improvements to metrics

gabegma:
* Add ability to disable trimming if needed - and track it in the output
* Fix trimming for models where the volume can be quite low (Voxtral)
* Fix interruption detection when there is a tool call
* Make sure metrics that failed because of a Memory Error will rerun (error used to be empty and wouldn't rerun)


tara-bogavelli (approved):

src/eva/metrics/processor.py
```diff
@@ -413,13 +413,16 @@ def _handle_audio_start(
...
-        # audio started in the current turn (not lingering from a previous turn's delayed delivery) and
-        # (b) no tool calls happened yet — tool calls mean the assistant is responding to the user's
-        # input, not barging in.
-        if state.user_audio_open and state.user_audio_started_in_turn and not state.assistant_processed_in_turn:
+        # Interruption: assistant audio_start overlaps an open user audio session. Flag the turn
+        # whenever there's overlap (validated against audio: real barge-ins occur even when a
+        # tool call has already fired this turn).
+        #
+        # `hold_turn` is gated on no-tool-call-yet: setting it after a tool call suppresses the
+        # next turn advance and cascades into incorrect turn boundaries downstream.
```
gabegma:
```suggestion
        # Interruption: assistant audio_start overlaps an open user audio session. Flag the turn
        # whenever there's overlap.
        # `hold_turn` if the assistant has not yet spoken.
```

PR #113, opened 2026-05-12
Add metric versionning

gabegma:
This will help when rerunning to track if all records have the latest version.


src/eva/utils/prompt_manager.py
```diff
@@ -81,6 +81,24 @@ def _load_from_directory(self, directory: Path) -> None:
...
+        value = self.prompts
+        for part in parts:
+            if not isinstance(value, dict):
+                raise KeyError(f"Invalid prompt path: {path} (stopped at {part})")
+            if part not in value:
+                raise KeyError(f"Prompt not found: {path} (missing key: {part})")
+            value = value[part]
+        if not isinstance(value, str):
+            raise ValueError(f"Prompt at {path} is not a string: {type(value)}")
+        return value
```
JosephMarinier, line 107: This is duplicated from the `get_prompt()` function below. Can we call `get_template()` insice `get_prompt()`?

src/eva/models/versioning.py
```diff
@@ -0,0 +1,24 @@
...
+import hashlib
+from contextvars import ContextVar
+
+_CURRENT_METRIC_VERSION: ContextVar[str | None] = ContextVar("current_metric_version", default=None)
+_CURRENT_PROMPT_HASH: ContextVar[str | None] = ContextVar("current_prompt_hash", default=None)
+
+
+def hash_prompt_template(template: str) -> str:
+    """Return sha256[:12] of an unrendered prompt template string."""
+    return hashlib.sha256(template.encode("utf-8")).hexdigest()[:12]
```
JosephMarinier: Detail: we can leave the default encoding, which is `"utf-8"`. That's what we do everywhere else in the repo where we call `str.encode()`.
```suggestion
    return hashlib.sha256(template.encode()).hexdigest()[:12]
```

tests/unit/metrics/test_metric_signatures.py
```diff
@@ -0,0 +1,83 @@
...
+
+import json
+from pathlib import Path
+
+import pytest
+
+from eva.metrics.signatures import compute_all_metric_signatures
+
+FIXTURE_PATH = Path(__file__).resolve().parents[3] / "tests" / "fixtures" / "metric_signatures.json"
+REGEN_HINT = "Run `python scripts/regen_metric_signatures.py` after bumping `version` on affected classes."
```
JosephMarinier, line 29: Do you think we should run `scripts/regen_metric_signatures.py` in pre-commit? It takes about 4 seconds to run, so I understand if we don't want to run it all the time.

JosephMarinier (approved):
Cool! Thanks for taking care of this!

The implementation seems a bit complex, but I don't expect this code to change often, so I think it's fine.

src/eva/utils/prompt_manager.py
JosephMarinier: Done in 6d66128b4a6801469099d59c9bf5c5a0794bb3f7

tests/unit/metrics/test_metric_signatures.py
gabegma: I would say no for now - unless we could run it only when metric files change?

gabegma:
> The implementation seems a bit complex, but I don't expect this code to change often, so I think it's fine.

I agree, it's somewhat ugly. Thanks for your cleanups!!

tests/unit/metrics/test_metric_signatures.py
JosephMarinier: Yes, we just need to list the files that can impact the signatures. Something like this?
```yaml
      - id: regen-metric-signatures
        name: Regenerate metric signatures
        entry: python3 scripts/regen_metric_signatures.py
        language: system
        pass_filenames: false
        files: |-
          (?x:
            ^configs/prompts/judge\.yaml$
            |
            ^scripts/regen_metric_signatures\.py$
            |
            ^src/eva/metrics/
            |
            ^src/eva/models/versioning\.py$
          )
```

src/eva/metrics/versioning.py
JosephMarinier, line 1: I don't think this file is in the right place. The other files in `src/eva/models/` define Pydantic **models**. Not this one. Did you mean to put it in `src/eva/metrics/` rather than `src/eva/models/`?

gabegma: Oh! Yes, my mistake, thank you!!

JosephMarinier: Done in 1b3602d

tests/unit/metrics/test_metric_signatures.py
JosephMarinier: Done in 0cb0f5e

src/eva/metrics/versioning.py
gabegma: Yay thank you!!

tests/unit/metrics/test_metric_signatures.py
gabegma: Amazing :)

PR #116, opened 2026-05-14
Website fixes: Perturbation significance asterisks; scope domain toggle to scatter plot

lindsaydbrin:
## Summary

Two related fixes to the Perturbations section of the leaderboard:

1. **Scope the domain toggle to the scatter plot only.** We'd intended to show pooled and not per-domain plots. `<Perturbations>` and `<PerturbationBarChart>` no longer take a `domain` prop and now call `getPertValue` with `'pooled'`.
2. **Fix significance asterisk placement and size.** `StarMark` now places markers just outside the CI cap in the bar's direction - above for positive deltas, below for negative - mirroring the implementation on the stats branch / for the paper. Font size now scales with `vb.width` (clamped to [7, 13]) so `***` always fits inside the bar. Asterisk `y` is also clamped inside the plot pixel range so it stays visible even if a CI cap is outside the [-0.5, 0.5] axis (which doesn't happen with current pooled data, but just in case).

Build artifacts under `docs/` are regenerated.

Follow-up: 
- *TL;DR Fixed a bug where zero values on the plots were ignored for indexing and asterisks were then positioned incorrectly.* 
- Details from Claude: Also fixes an index-drift bug where recharts' Bar drops zero-dimension rectangles before building the LabelList context, causing the content callback's index to point into a filtered array. Switched to valueAccessor so we read each row's data straight off the recharts entry.

## Test plan

- [x] `npm run build` — clean (vite build, tsc -b)
- [x] `npx eslint` on touched files — clean
- [x] `npm run dev` — visually verified asterisks are attached to their bars (above for positive, below for negative) and shrink to fit narrow bars
- [x] Toggling Pooled / CSM / ITSM / HR re-renders the scatter plot only; Perturbations section stays put


lindsaydbrin:
Screenshot 2026-05-14 at 11 47 59 AM

Screenshot 2026-05-14 at 11 44 17 AM

tara-bogavelli:
I think you need to pull main and rebuild since you have some merge conflicts

tara-bogavelli (approved):

PR #119, opened 2026-05-15
add config app

raghavm243512:
add config app to manage ever growing environment variables
As we grow the .env.example, just need to 
1. place them in the right section
2. tag with a couple metadata comments

Regular comments are still supported in the .env.example file

Core logic:
Read the .env.example as a schema to define expected variables, datatypes, & organization.
Read the .env for existing values, and fill them in

few different options to save, view, copy etc the generated env

use `streamlit run apps/config_editor.py`


apps/config_io.py
```diff
@@ -0,0 +1,391 @@
...
+
+    # ── back-compat shim so old tests still compile ───────────────────────
+    @property
+    def specs(self) -> list[AnnotatedVar]:
+        return self.vars
+
+
+def _is_section_rule(line: str) -> bool:
+    s = line.strip()
+    return s.startswith("#") and set(s.lstrip("#").strip()) == {"="} and len(s) > 5
```
katstankiewicz: can we use regex here, like
`return bool(re.match(r"^#+\s*={3,}\s*$", s))`

(^ — start of string
#+ — one or more # characters
\s* — optional whitespace
={3,} — three or more = characters
\s* — optional trailing whitespace
$ — end of string)

apps/config_editor.py
```diff
@@ -0,0 +1,655 @@
...
+    st.markdown(f"**{name}**" + (f" — {info}" if info else ""))
+    raw_key = f"raw_{name}"
+    if raw_key not in st.session_state:
+        st.session_state[raw_key] = json.dumps(current, indent=2) if current else ""
+
+    rows = [{"key": k, "value": _scalar_to_str(v)} for k, v in current.items()] or [{"key": "", "value": ""}]
+    edited = st.data_editor(
+        rows,
+        num_rows="dynamic",
+        use_container_width=True,
```
katstankiewicz: can you change these to `width="stretch"`

I saw this message while testing:
```
2026-05-19 10:20:34.279 Please replace `use_container_width` with `width`.

`use_container_width` will be removed after 2025-12-31.

For `use_container_width=True`, use `width='stretch'`.
```

katstankiewicz (approved):

PR #120, opened 2026-05-16
Improve speech fidelity

gabegma:
* Fix some issues we were seeing with false positives, and add submetrics for speech fidelity.


src/eva/metrics/experience/conciseness.py
```diff
@@ -70,21 +70,13 @@ def build_sub_metrics(
...
         """Surface one sub-metric per failure mode, rate = flagged turns / rated turns."""
         rated_turn_ids = [tid for tid, r in per_turn_ratings.items() if r is not None]
-        num_rated = len(rated_turn_ids)
-        if num_rated == 0:
-            return None
-
-        sub_metrics: dict[str, MetricScore] = {}
-        for mode in _CONCISENESS_FAILURE_MODES:
-            flagged_ids = [
-                tid for tid in rated_turn_ids if mode in (per_turn_extra.get(tid, {}).get("failure_modes") or [])
```
JosephMarinier, line 80: Was there already code for parsing `"failure_modes"`, but it wasn't asked in the prompt? Or am I missing something?

src/eva/metrics/speech_fidelity_base.py
```diff
@@ -108,17 +109,22 @@ async def compute(self, context: MetricContext) -> MetricScore:
                 rating = response_item.get("rating")
                 transcript = response_item.get("transcript")
                 explanation = response_item.get("explanation", "")
+                failure_modes = response_item.get("failure_modes") or []
+                if not isinstance(failure_modes, list):
+                    failure_modes = []
```
JosephMarinier: Small detail, but the `or []` is redundant with the `if` below. We can remove it:
```suggestion
                failure_modes = response_item.get("failure_modes")
                if not isinstance(failure_modes, list):
                    failure_modes = []
```
Or if the `if` below is overkill (if we don't need to handle `"failure_modes": 42`), we can remove it and keep `or []`:
```suggestion
                failure_modes = response_item.get("failure_modes") or []
```

JosephMarinier (approved):

src/eva/metrics/experience/conciseness.py
gabegma: This is in the concisness file - so we reused the same code for both speech fidelity and conciseness now, since they now both use failure_modes - speech fidelity didn't before, but conciseness did. Is that your question?

JosephMarinier: Yes! Thank you!

PR #121, opened 2026-05-18
Multilingual support

raghavm243512:
initial multilingual version

Easily extendable to many language using the add_culture_data script. This will do translation, gender consistent naming, suggest names, extend data, etc. So if anyone wants to run a language not committed in EVA data, it is trivially easy to do so.
Readme section showing basic of adding a language. 

This adds:
Multilingual data schema and content (initial utterances, system prompt, name aliases)
multilingual support in code
Prompt updates to support multi languages
Script to "add a language" with high degree of automation
WER metric normalization rules, dynamically set per language and creatable via LLM through adding script
Automatic .env.example adjustments (maintains config app accuracy)

Still TODO:
- [ ] Currencies
- [x] Phone numbers (airline only problem)
- [x] Actually committing the translations (didn't want to burn credits until finalized)
- [ ] Analysis
- [ ] Testing a large variety of models to ensure they actually get the language code they expected (es-MX vs es, for example)


src/eva/utils/culture.py
```diff
@@ -0,0 +1,153 @@
...
+def _load_initial_messages() -> dict[str, str]:
+    if not INITIAL_MESSAGES_PATH.exists():
+        return {}
+    return yaml.safe_load(INITIAL_MESSAGES_PATH.read_text(encoding="utf-8")) or {}
+
+
+def get_initial_message(language: str) -> str:
+    """Return the assistant's opening line for ``language``.
+
+    Falls back to English. Raises if even English is missing (data quality).
```
katstankiewicz:
```suggestion
    Falls back to English. Raises if even English is missing.
```

scripts/run_text_only.py
```diff
@@ -510,7 +524,13 @@ async def run_record(
     logger.info(f"Text-only test: record {record.id} | model={llm_model} | max_turns={max_turns}")
     logger.info(f"Metrics: {', '.join(requested_metrics)}")
 
-    user_message = record.user_goal["starting_utterance"]
+    user_message = resolve_user_goal(
+        record.user_goal,
+        record.culture_overrides,
+        os.getenv("EVA_LANGUAGE", "en"),
```
katstankiewicz:
```suggestion
        language,
```

scripts/run_text_only.py
```diff
@@ -173,7 +174,13 @@ def resolve_paths(domain: str) -> tuple[Path, Path, Path]:
 def build_user_sim_prompt(record: EvaluationRecord) -> str:
     """Build the user-simulator system prompt from the record's goal and persona."""
     pm = PromptManager()
-    goal = record.user_goal
+    goal = resolve_user_goal(
+        record.user_goal,
+        record.culture_overrides,
+        os.getenv("EVA_LANGUAGE", "en"),
```
katstankiewicz:
```suggestion
        language,
```
and add `language` to `build_user_sim_prompt`

src/eva/assistant/pipeline/services.py
```diff
@@ -468,6 +477,7 @@ def create_realtime_llm_service(
                 ),
                 audit_log=audit_log,
                 api_key=params["api_key"],
+                base_url=params.get("url", ""),
```
```suggestion
```
we don't need a url for openai services

scripts/add_culture_data.py
```diff
@@ -0,0 +1,436 @@
...
+from pathlib import Path
+from typing import Any
+
+import yaml
+from dotenv import load_dotenv
+
+from eva.utils.culture import FIRST_NAME_PLACEHOLDER, LAST_NAME_PLACEHOLDER
+from eva.utils.json_utils import extract_and_load_json
+from eva.utils.llm_client import LLMClient
+from eva.utils.logging import get_logger
```
katstankiewicz:
```suggestion
from eva.utils.logging import get_logger, setup_logging
```

scripts/add_culture_data.py
```diff
@@ -0,0 +1,436 @@
...
+from eva.utils.router import init
+
+logger = get_logger(__name__)
```
katstankiewicz, line 99: adds logging (previously was only printing warning level or above)
```suggestion
setup_logging()
logger = get_logger(__name__)
```

scripts/add_culture_data.py
```diff
@@ -0,0 +1,436 @@
...
+    # 2. Translate utterances in batch.
+    if to_translate_text:
+        logger.info(f"Translating {len(to_translate_text)} utterances to {language_name}")
+        translated = await _translate_utterances(to_translate_text, language_name, llm)
+        for idx, t in zip(to_translate_idx, translated, strict=True):
+            records[idx]["starting_utterances"][language] = t
+
+    # 3. Write back atomically.
+    if dry_run:
+        logger.info(f"[dry-run] would update {dataset_path} ({len(records)} records)")
```
katstankiewicz:
```suggestion
        logger.info(f"[dry-run] would update {dataset_path} ({len(target_ids)} records)")
```

scripts/add_culture_data.py
```diff
@@ -0,0 +1,436 @@
...
+    if not isinstance(rom, list) or len(rom) != len(flat_native):
+        raise ValueError(f"Expected {len(flat_native)} romanized names, got: {rom!r}")
+    # Reconstruct: ASCII half unchanged, native half romanized — indices stay aligned.
+    return {key: names[key][:BUCKET_SIZE] + rom[spans[key][0] : spans[key][1]] for key in spans}
+
+
+def _load_names_file(path: Path) -> dict[str, list[str]]:
+    data = json.loads(path.read_text(encoding="utf-8"))
+    for key in ("male_first", "female_first", "last"):
+        if not isinstance(data.get(key), list) or not data[key]:
```
katstankiewicz: should this check that there are enough names as well? ie `BUCKET_SIZE`?

katstankiewicz (commented):
can you also add `ensure_ascii=False` to AuditLog `save()`

src/eva/assistant/pipeline/services.py
raghavm243512: I added this because I was testing a side project (self hosted S2S agent). With any luck we will need this so we can leave or remove up to you

configs/agents/initial_messages.yaml
```diff
@@ -0,0 +1,6 @@
+de: Hallo! Wie kann ich dir heute helfen?
+en: Hello! How can I help you today?
+es: ¡Hola! ¿En qué puedo ayudarte hoy?
+fr: Bonjour ! Comment puis-je vous aider aujourd'hui ?
```
fanny-riols: We don't have fq?

configs/prompts/judge.yaml
```diff
@@ -333,12 +334,12 @@ judge:
       - Using parameter values explicitly stated by the user
       - Using parameter values returned by a prior tool response (e.g., reusing an ID or record returned by an earlier lookup in a subsequent call)
       - Using reasonable defaults that are standard for the tool (e.g., a date format conversion)
-      - Standard domain mappings from user-stated information (e.g., "Chicago O'Hare" → "ORD", "Miami" → "MIA"; or other unambiguous geographic, enterprise, or industry-standard mappings present in the agent's domain) — unambiguous mappings are considered grounded
+      - Standard domain mappings from user-stated information (e.g., "Chicago O'Hare" → "ORD"; "sore throat and fever" → ICD-10 code "J06.9"; or other unambiguous geographic, enterprise, or industry-standard mappings present in the agent's domain) — unambiguous mappings are considered grounded
```
fanny-riols: I'm not sure I understand that part, where is it coming from?

configs/prompts/judge.yaml
```diff
@@ -578,6 +581,9 @@ judge:
        **Binary Rating:**
       - **1 (Clean)**: The user's behavior did not corrupt the agent evaluation. None of the corruption types above occurred. Minor deviations from the user's instructions that did not affect database state are acceptable.
       - **0 (Corrupted)**: One or more corruption types occurred — the user's behavior caused the agent to be evaluated against an incorrect database state.
+
+      **Language:**
+      - Always provide analysis in English
```
fanny-riols:
```suggestion
      - Always provide analysis in English.
```

configs/prompts/judge.yaml
```diff
@@ -694,8 +703,8 @@ judge:
           "turns": [
             {{
               "turn_id": <int: the turn number from the Intended Turns>,
-              "transcript": <string: your transcription of the audio for this turn, use only the audio for this not the intended text>
-              "explanation": "<string: 1-3 sentence analysis of fidelity for this turn, citing specific intended vs actual mismatches, noting any regions skipped due to interruption flags>",
+              "transcript": <string: your transcription of the audio for this turn using the appropriate script for the language spoken, use only the audio for this not the intended text>
```
fanny-riols: I'm confused on this addition, what do we mean by "appropriate script"?

configs/prompts/simulation.yaml
```diff
@@ -29,11 +29,11 @@ agent:
     
     ### Clarity When Spoken Aloud
     - Spell out acronyms and abbreviations in full (say "as soon as possible" not "ASAP", "by the way" not "BTW")
-    - Express numbers in spoken form appropriate to context:
-      * Dates: "January 15th, 2024" not "1/15/2024"
-      * Times: "three thirty PM" not "3:30 PM"
-      * Quantities: "twenty dollars" not "$20"
-      * Years: "twenty twenty-four" not "2024"
+    - Express numbers in linguistically/culturally appropriate spoken form for the conversation language:
+      * Dates: spoken form matching local convention (e.g. English: "January 15th, 2024"; French: "le quinze janvier deux mille vingt-quatre") not "1/15/2024"
```
fanny-riols:
```suggestion
      * Dates: spoken form matching local convention (e.g. English: "January 15th, 2024"; French: "le quinze janvier deux mille vingt-quatre"; not "1/15/2024")
```

configs/prompts/simulation.yaml
```diff
@@ -29,11 +29,11 @@ agent:
...
+      * Times: spoken form (e.g. English: "three thirty PM"; German: "fünfzehn Uhr dreißig") not "3:30 PM"
```
fanny-riols:
```suggestion
      * Times: spoken form (e.g. English: "three thirty PM"; German: "fünfzehn Uhr dreißig"; not "3:30 PM")
```

configs/prompts/simulation.yaml
```diff
@@ -29,11 +29,11 @@ agent:
...
+      * Quantities: spoken form (e.g. English: "twenty dollars"; Korean: "이십 달러") not "$20"
```
fanny-riols:
```suggestion
      * Quantities: spoken form (e.g. English: "twenty dollars"; Korean: "이십 달러"; not "$20")
```

configs/prompts/simulation.yaml
```diff
@@ -29,11 +29,11 @@ agent:
...
+      * Years: spoken form appropriate to the language, not "2024"
```
fanny-riols:
```suggestion
      * Years: spoken form appropriate to the language (e.g. English: "twenty twenty-four"; not "2024")
```

configs/prompts/simulation.yaml
```diff
@@ -87,11 +87,11 @@ audio_llm_agent:
...
```
fanny-riols:
```suggestion
      * Dates: spoken form matching local convention (e.g. English: "January 15th, 2024"; French: "le quinze janvier deux mille vingt-quatre"; not "1/15/2024")
      * Times: spoken form (e.g. English: "three thirty PM"; German: "fünfzehn Uhr dreißig"; not "3:30 PM")
      * Quantities: spoken form (e.g. English: "twenty dollars"; Korean: "이십 달러"; not "$20")
      * Years: spoken form appropriate to the language (e.g. English: "twenty twenty-four"; not "2024")
```

configs/prompts/simulation.yaml
```diff
@@ -150,11 +150,11 @@ realtime_agent:
...
```
fanny-riols:
```suggestion
      * Dates: spoken form matching local convention (e.g. English: "January 15th, 2024"; French: "le quinze janvier deux mille vingt-quatre"; not "1/15/2024")
      * Times: spoken form (e.g. English: "three thirty PM"; German: "fünfzehn Uhr dreißig"; not "3:30 PM")
      * Quantities: spoken form (e.g. English: "twenty dollars"; Korean: "이십 달러"; not "$20")
      * Years: spoken form appropriate to the language ("twenty twenty-four" not "2024")
```

data/itsm_scenarios/1.json
```diff
@@ -405,8 +405,37 @@
...
         "name": "Downtown Engineering Center",
         "name_aliases": [
-          "Downtown",
-          "Engineering Center"
+          "downtown",
+          "engineering center",
+          "centre d’ingénierie du centre-ville",
+          "ingénierie centre-ville",
+          "bâtiment ingénierie centre-ville",
+          "le centre d’ingénierie en ville",
```
fanny-riols: I would drop this one as it wouldn't be accurate (translating downtown to town)
```suggestion
```

data/itsm_scenarios/1.json
```diff
@@ -405,8 +405,37 @@
       "BLD3": {
         "name": "Downtown Engineering Center",
         "name_aliases": [
-          "Downtown",
-          "Engineering Center"
+          "downtown",
+          "engineering center",
```
fanny-riols: Since we have this alias in English, I think we should also add the translations for these two in other languages.

Also, can you sort translation by language please?

data/itsm_scenarios/1.json
```diff
@@ -468,8 +497,34 @@
       "PZA": {
         "name": "Garage A",
         "name_aliases": [
-          "A Garage",
-          "Main Garage"
+          "a garage",
+          "main garage",
+          "garage a",
```
fanny-riols: Wouldn't that match with the name already?

data/itsm_scenarios/10.json
```diff
@@ -405,8 +405,37 @@
       "BLD3": {
         "name": "Downtown Engineering Center",
         "name_aliases": [
-          "Downtown",
-          "Engineering Center"
+          "downtown",
```
fanny-riols: Also, since we have all these alias in each file, shouldn't we put all the aliases in a separate common file?

configs/prompts/judge.yaml
raghavm243512: just an example of medical stuff. I know we are doing HR and not actual patient care but I figure general examples are always good to include even if not about each domain directly

configs/agents/initial_messages.yaml
raghavm243512: easy to add, will do. The existing languages are just things I saved from testing because why waste them

configs/prompts/judge.yaml
raghavm243512: should write `你好` not `ni hao`

data/itsm_scenarios/1.json
raghavm243512: Not sure what you mean

data/itsm_scenarios/1.json
raghavm243512: What do you mean by the first part? there are translations in other languages?

data/itsm_scenarios/1.json
fanny-riols: The name on row 498 is `Garage A`, and here we add the alias `garage a`. Is the match check case-sensitive? If not, the alias is the same as the name.

data/itsm_scenarios/1.json
fanny-riols: In French, `downtown` would be translated to `centre ville`, which is not part of the list below. There are some variants like `ingénierie centre-ville`, but this means `downtown engineering`, not `downtown` alone.

data/itsm_scenarios/1.json
raghavm243512: Name check isn't case insensitive. This is also pretty small but realistically the name check isn't likely to hit super often and alias is the real important part here

data/itsm_scenarios/10.json
raghavm243512: Definitely can do soon. Can publish first though (blog timing) since this is a nicety and less functional

src/eva/utils/wer_normalization/configs/fr-CA.json
```diff
@@ -0,0 +1,258 @@
...
+      "ère"
+    ],
+    "second": [
+      2,
+      "d"
+    ],
+    "seconde": [
+      2,
+      "de"
+    ],
```
JosephMarinier, line 50: I don't think I ever saw "2d" and "2de" with that meaning. I would assume "2d" means two dimensions rather than "second". Also, could that confuse "seconde" (the unit of time) with "2de"?

src/eva/utils/wer_normalization/configs/en.json
```diff
@@ -0,0 +1,480 @@
...
+      "th"
+    ],
+    "first": [
+      1,
+      "st"
+    ],
+    "second": [
+      2,
+      "nd"
+    ],
```
JosephMarinier, line 118: How is this mapping used? Could that confuse "second" (the unit of time) with "2nd"?

src/eva/utils/wer_normalization/configs/fr-CA.json
```diff
@@ -0,0 +1,258 @@
...
+      "e"
+    ],
+    "octantième": [
+      80,
+      "e"
+    ],
+    "nonantième": [
+      90,
+      "e"
+    ]
```
JosephMarinier, line 168: These are used in Switzerland, Belgium, and Luxembourg, but in France and Canada, we say "soixante-dixième", etc. I think that is covered by "soixante" in "tens" plus "dixième" in "cardinals_suffixed", right?
```suggestion
```

JosephMarinier: That said, it might not hurt to keep them, even if they are not supposed to be used.

src/eva/utils/wer_normalization/configs/fr-CA.json
JosephMarinier: I checked, and these are official abbreviations, but I didn't know them, and as I said, I have never seen them.

src/eva/utils/wer_normalization/configs/fr.json
```diff
@@ -0,0 +1,261 @@
...
+      "ème"
+    ],
+    "dix-huitième": [
+      18,
+      "ème"
+    ],
+    "dix-neuvième": [
+      19,
+      "ème"
+    ]
```
JosephMarinier, line 122: In both France and Canada, "ème" is technically incorrect. Instead, we are supposed to write "e" (like you did in `fr-CA.json` 👍).

That said, it is a very common mistake. Maybe we should support both? For sure, we should at least support the correct "e".

raghavm243512: The json configs for WER normalization are AI generated, then run through some automated tests (Ask LLM to make string versions of the numbers, and we attempt to parse it back)

Now that I think about it I should make that invocation separate to reduce test case bias in generation. 

But the specifics of the file I can't really give meaningful insight on since I don't know the language. Feel free to make edits if there is a clear gap

src/eva/utils/wer_normalization/configs/en.json
raghavm243512: it absolutely would, however it wouldn't impact calculation of WER

tests/unit/utils/test_multilingual_normalization.py
```diff
@@ -0,0 +1,279 @@
...
+    KoreanTextNormalizer,
+)
+from eva.utils.wer_normalization.wer_utils import normalize_text
+
+# ===========================================================================
+# French — number equivalence (digits ↔ spelled-out)
+# ===========================================================================
+
+
+class TestFrenchNumbers:
```
gabegma: This test is kind of redundant with the one in test_wer_utils, but with different values? Could we clean that up?

tests/unit/utils/test_wer_utils.py
```diff
@@ -83,7 +135,7 @@ def test_21st_not_converted_to_saint(self):
         ],
     )
     def test_ordinals_preserved(self, text: str, expected: str):
-        assert normalize_text(text) == expected
+        assert normalize_text(text, Language.FR.value) == expected
```
gabegma: Shouldn't this be English?

src/eva/utils/wer_normalization/configs/README.md
```diff
@@ -0,0 +1,336 @@
+# WER Normalizer Config Reference
```
gabegma, line 1: Was this taken from another repo? If yes, we should cite here.

configs/agents/airline_agent.yaml
```diff
@@ -5,8 +5,16 @@ role: You are a voice agent for SkyWay Airlines handling inbound calls for fligh
...
+
   Every call begins with authentication. Ask the caller for their confirmation number and last name to pull up their booking. Confirm you have the correct reservation before proceeding.
+
+  ### Name Disambiguation
+
+  If a lookup fails or returns no match, use the following escalation path:
+
+  1. Ask the caller to confirm or spell their last name — once, clearly.
+  2. If the spelling is confirmed correct and the lookup still fails, ask for a secondary signal: the departure date, origin city, or destination.
+  3. If the caller has a very common last name (one that alone is insufficient to narrow results), lead with the confirmation number as the primary key and treat the last name as a secondary check, not the other way around.
```
gabegma: I don't understand these? For 2 - what would it change to ask for the departure date or the destination? The tools don't allow for this, right? I also don't understand what 3 means here.

src/eva/utils/wer_normalization/configs/README.md
raghavm243512: No, this is intended as a helper for people to do manual correction if the LLM generation isn't good enough

scripts/add_culture_data.py
```diff
@@ -0,0 +1,1548 @@
...
+        f"For {language_name} ({language}), what is the standard mobile phone number format?\n"
+        "Make sure you're aware that the region is the defining factor (eg, pt-BR would use Brazilian phone numbers).\n"
+        "Return JSON with:\n"
+        "  calling_code: string (country calling code, no leading +)\n"
+        "  mobile_format: a format string where X = random digit, fixed digits are written literally,\n"
+        "    and separators (spaces, hyphens, dots) are included as-is.\n"
+        'Example for France (+33 6 XX XX XX XX): {"calling_code": "33", "mobile_format": "6 XX XX XX XX"}\n'
+        'Example for UK (+44 7XXX XXXXXX): {"calling_code": "44", "mobile_format": "7XXX XXXXXX"}\n'
+        'Example for US (+1 NXX-NXX-XXXX): {"calling_code": "1", "mobile_format": "XXX-XXX-XXXX"}\n'
+        'Example for Germany (+49 151 12345678): {"calling_code": "49", "mobile_format": "151 XXXXXXXX"}\n'
```
JosephMarinier: At this point, we're giving the languages we care about as examples, so the LLM call seems a bit overkill. Have you considered making this more deterministic by hardcoding a map of phone number formats?

configs/agents/airline_agent.yaml
```diff
@@ -5,8 +5,15 @@ role: You are a voice agent for SkyWay Airlines handling inbound calls for fligh
...
-  
+
   Every call begins with authentication. Ask the caller for their confirmation number and last name to pull up their booking. Confirm you have the correct reservation before proceeding.
+
+  ### Name Disambiguation
+
+  If a lookup fails or returns no match, use the following escalation path:
+
+  1. Ask the caller to confirm or spell their last name — once, clearly.
+  2. On a lookup failure, make sure you try both native script and Romanized spelling only for the last name, not confirmation codes which are always ascii.
```
gabegma: I'm still a bit nervous about this new section, which might lead the agent to no longer try to disambiguate the confirmation code, which is often the wrong one of the 2, at least in English. I would suggest perhaps something like the following?

```suggestion
  ### Disambiguation

  If a lookup fails or returns no match, use the following escalation path:

  1. Ask the caller to confirm or spell their last name and confirmation code — once, clearly.
  2. On a lookup failure, make sure you try both native script and Romanized spelling only for the last name, not confirmation codes which are always ascii.
```

configs/agents/language_addenda.yaml
```diff
@@ -0,0 +1,16 @@
...
+  response which says 'downtown', this should be translated. Distinct item names (e.g.
+  'IntelliJ') should be kept in their original form.
+fr: Always respond to the user in European French, regardless of the instructions
+  given or tool outputs received. However, tool calls and tool names must always be
+  done using ascii characters, except parameters like people's first or last names
+  which may be in non-ascii, native script. You may need to try both scripts when
+  looking up by name. All translatable values should be translated when talking to
+  the user. For example, if you are telling the user about a location from a tool
+  response which says 'downtown', this should be translated. Distinct item names (e.g.
+  'IntelliJ') should be kept in their original form.
```
gabegma: This looks like it would be the same for all languages except for the language name? Instead of duplicating this so many times, could we simply have it in the Python file and just pass the language as a variable?

configs/prompts/judge.yaml
```diff
@@ -82,6 +82,7 @@ judge:
 
       **verbosity_or_filler**
       Contains unnecessary wording, repetition within the same turn, hedging, or explanation beyond what the context requires.
+      Exception: do NOT flag following of cultural norms, such as obligatory politeness structures, as filler (e.g. in Japanese and Korean: keigo, honorifics, set opening/closing phrases).
```
gabegma: This reads a bit weirdly to me.
```suggestion
      Exception: do NOT flag following of cultural norms, such as obligatory politeness structures (e.g. in Japanese and Korean: keigo, honorifics, set opening/closing phrases).
```

configs/prompts/judge.yaml
gabegma: In what dataset do we have ` "sore throat and fever" → ICD-10 code "J06.9"`? Is this an American thing?

configs/prompts/judge.yaml
```diff
@@ -333,12 +334,12 @@ judge:
...
-      - Trying valid variations of a user-stated identifier or name when a prior read-only lookup failed (e.g., swapping commonly confused letters like "v"/"z" or "b"/"d", reordering characters, trying alternative spellings) — this is proper disambiguation behavior, not fabrication, since the variation is grounded in the user's original utterance
+      - Trying valid variations of a user-stated identifier or name when a prior read-only lookup failed (e.g., swapping commonly confused letters, reordering characters, trying alternative spellings) — this is proper disambiguation behavior, not fabrication, since the variation is grounded in the user's original utterance. Which letters are phonetically confusable is language-dependent: in English "v"/"z" or "b"/"d" are classic pairs; in Korean, M is named "엠" and N is named "엔" — nearly identical in fast or phone-quality audio; other languages have their own similar-sounding pairs. The agent should apply the same disambiguation logic regardless of which language's phonetic naming convention the user is using.
```
gabegma: What does this mean? `The agent should apply the same disambiguation logic regardless of which language's phonetic naming convention the user is using.`

configs/prompts/judge.yaml
```diff
@@ -511,6 +513,7 @@ judge:
       The conversation ended due to: {conversation_end}
         
       ## User Simulator Instructions
+      The user in this conversation was expected to communicate in {language_display_name} throughout.
```
gabegma, line 516: Could we add a corruption type specifically for issues related to users who do not speak the correct language? This would help the LLM in detecting those, and would help to see how much of an issue this is.

configs/prompts/judge.yaml
```diff
@@ -578,6 +581,9 @@ judge:
        **Binary Rating:**
       - **1 (Clean)**: The user's behavior did not corrupt the agent evaluation. None of the corruption types above occurred. Minor deviations from the user's instructions that did not affect database state are acceptable.
       - **0 (Corrupted)**: One or more corruption types occurred — the user's behavior caused the agent to be evaluated against an incorrect database state.
+
+      **Language:**
+      - Always provide analysis in English.
       
       Respond in JSON format:
         {{
```
gabegma: I would keep consistent with the other prompts and not add a new section
```suggestion
     
     Respond in JSON format:
     Provide analysis in English.
       {{
```

configs/prompts/judge.yaml
```diff
@@ -677,12 +683,15 @@ judge:
...
         - Slight pacing or prosody differences
         - Non-spoken tags: [slow], [firm], [annoyed], and all interruption tags listed above
         - Words in regions flagged by interruption tags as likely not spoken
+        - Phonetic spellings of Latin/Roman letters in non-Latin-script languages (e.g., Korean "에이" for "A", "비" for "B") — these are the correct spoken form of those letters in that language and are not errors
+        - Regional and language-structural number variants that represent the same value — accept any valid spoken form; many languages have systematic surface differences that are not errors: regional variants (French "septante" vs "soixante-dix" for 70), reversed digit order (German spoken form reverses tens and units: "vierundzwanzig" = four-and-twenty = 24), dual counting systems (Korean native/Sino-Korean, Japanese on/kun reading), large-unit grouping differences (Japanese 万-based system: "ichi-man" = 10,000), or grammatical agreement variants (Spanish "veintiún" vs "veintiuna")
```
gabegma: I think it should be an error to use septante for French Canadian - it's not something we use at all, and I wouldn't understand it - it doesn't seem to be common in France either.  I would give the following example: "dix-huit cents trois vs mille huit cents trois".  Could we validate the other examples with fluent speakers of these languages?

configs/prompts/judge.yaml
gabegma: the appropriate script?

configs/prompts/judge.yaml
```diff
@@ -694,8 +703,8 @@ judge:
           "turns": [
             {{
               "turn_id": <int: the turn number from the Intended Turns>,
-              "transcript": <string: your transcription of the audio for this turn, use only the audio for this not the intended text>
-              "explanation": "<string: 1-3 sentence analysis of fidelity for this turn, citing specific intended vs actual mismatches, noting any regions skipped due to interruption flags>",
+              "transcript": <string: your transcription of the audio for this turn using the appropriate script for the language spoken, use only the audio for this not the intended text>
+              "explanation": "<string: 1-3 sentence analysis in English of fidelity for this turn, citing specific intended vs actual mismatches, noting any regions skipped due to interruption flags>",
```
gabegma:
```suggestion
              "explanation": "<string: 1-3 sentence analysis in English of fidelity for this turn, citing specific intended vs actual mismatches in the original language, and noting any regions skipped due to interruption flags, if applicable>",
```

configs/prompts/judge.yaml
```diff
@@ -767,9 +776,9 @@ judge:
           "turns": [
             {{
               "turn_id": <int: the turn number from the Conversation Trace>,
-              "transcript": <string: your transcription of the audio for this turn, use only the audio for this not the conversation trace>,
+              "transcript": <string: your transcription of the audio for this turn using the appropriate script for the language spoken, use only the audio for this not the conversation trace>,
```
gabegma: Same question for script?

configs/prompts/judge.yaml
```diff
@@ -898,7 +911,16 @@ judge:
...
+      - Number surface variants: Many languages have systematic spoken-form differences that do not change the underlying value and must not be penalized. Accept any of the following as correct:
+        - Reversed digit order in spoken form (e.g., German "vierundzwanzig" = four-and-twenty = 24 — this is the standard spoken form, not an error)
+        - Grammatical gender/agreement variants (e.g., Spanish "veintiún" vs "veintiuna" for 21)
+        - Large-unit grouping differences (e.g., Japanese 万-system: "ichi-man go-sen" = 15,000; accept either Western or Japanese grouping if the value matches)
+        Note — context-dependent number systems: Some languages require a specific number system depending on context and using the wrong one is an error. In Korean, hours are counted using the native Korean system (하나, 둘…) while minutes, prices, and most other quantities use Sino-Korean (일, 이…). In Japanese, counters (助数詞) impose specific reading conventions. Only flag these as wrong if the system used is clearly inappropriate for the context (e.g., a price stated in native Korean numbers).
```
gabegma, line 932: Was this validated with Japanese and Korean speakers?

scripts/add_culture_data.py
raghavm243512: I mean I could, but then the script loses its somewhat universal usability. We still care about polish, Spanish, Korean, Japanese, Hindi, Chinese. These don't have any examples so I think this is fine. But importantly, this is so anyone can get started on EVA multilang without needing us to publish their language of choice

configs/agents/language_addenda.yaml
raghavm243512: yea will do. I left it like this to leave it editable to the user discretion, but that is likely way overkill

configs/prompts/judge.yaml
raghavm243512: script here means not using "ni hao" but the actual Chinese characters

configs/prompts/judge.yaml
raghavm243512: The user's information might be in native script but it could also be stored in latin script. This is saying the agent should attempt the same strategies to disambiguate the name even in non latin scripts. It's probably unneeded though as I think it may be implied in the rest of the instruction

configs/prompts/judge.yaml
raghavm243512: I included a general example for code mapping from healthcare. Fanny had a similar issue with it so I just changed to something more contextually relevant

configs/prompts/judge.yaml
raghavm243512: septante is apparently used in Belgium and Switzerland. But for Canadian I can make some additional clarity

configs/prompts/judge.yaml
gabegma: Are you sure this is clear? Fanny was also confused here.

data/airline_scenarios/1.2.2.json
```diff
@@ -8,10 +8,10 @@
         {
           "passenger_id": "PAX001",
           "first_name": "Rachel",
```
gabegma: Is it on purpose that Rachel is kept for this one?

data/airline_scenarios/2.2.5.json
```diff
@@ -8,10 +8,10 @@
         {
           "passenger_id": "PAX001",
           "first_name": "Jordan",
```
gabegma: Same for Jordan here

data/itsm_scenarios/1.json
```diff
@@ -468,8 +482,18 @@
       "PZA": {
         "name": "Garage A",
         "name_aliases": [
-          "A Garage",
-          "Main Garage"
+          "a garage",
```
gabegma: I don't understand `name_aliases` - how is this used, and why is there no indication that this is French here?

gabegma: I see ok, for fuzzy matching - so we'll put all languages together. But aren't we expecting the agent to use the English term when making the tool call?

configs/prompts/judge.yaml
gabegma: yeah exactly, which is neither FR-CA nor French from France, the 2 "dialects" the agent should be speaking. This is why I would drop this example. The other example I gave works for both French Canada and France.

data/itsm_scenarios/1.json
raghavm243512: No specific expectation of that. The aliasing is enabled in English as well, so to allow the same "human lingo" to work for matching, we need the multi language fuzzy match

data/itsm_scenarios/1.json
```diff
@@ -525,7 +549,11 @@
         "slack"
       ],
       "requires_manager_approval": false,
-      "vendor": "Slack Technologies"
+      "vendor": "Slack Technologies",
```
gabegma: Is it normal that some fields like this one don't have `name_aliases` alone? I don't understand why this one doesn't, but JetBrains does below.

data/airline_scenarios/7.4.1.json
```diff
@@ -8,10 +8,10 @@
         {
           "passenger_id": "PAX001",
           "first_name": "Mina",
-          "last_name": "Park",
+          "last_name": "<LAST_NAME>",
           "ticket_number": "1801234567890",
-          "email": "mina.park@example.com",
-          "phone": "+1-415-555-0134",
+          "email": "mina.<LAST_NAME_ROMANIZED>@example.com",
```
gabegma: Mina here

data/airline_scenarios/4.1.2.json
```diff
@@ -30,12 +30,12 @@
       "passengers": [
         {
           "elite_status": null,
-          "email": "jordan.hill@example.com",
+          "email": "jordan.<LAST_NAME_ROMANIZED>@example.com",
```
gabegma: And another Jordan here.

data/itsm_scenarios/10.json
gabegma: I'm also not so convinced about the fuzzy matching approach - I wonder if we could change our tools so that the agent always uses English. I find it's inconvenient to grow this list as we add new languages, and I'm not sure how robust it will be. For example, in French, there are many different combinations that could be valid, even some not listed here.

data/itsm_scenarios/22.json
```diff
@@ -525,10 +548,15 @@
       ],
       "name": "JetBrains IntelliJ IDEA",
       "name_aliases": [
-        "IntelliJ",
-        "IDEA"
+        "intellij",
+        "idea"
```
gabegma: Sometimes `idea` is there, sometimes it's not. Sometimes I also see `intellij idea` - I think we should really save them in one place to avoid this issue.

configs/prompts/simulation.yaml
```diff
@@ -275,16 +275,18 @@ user_simulator:
...
-    - Code: "A B C one two three"
+    - Phone: say each digit individually in a linguistically/culturally appropriate way (e.g. English: "five five five... one two three... four five six seven"; Korean: "공 일 공... 일 이 삼... 사 오 육 칠")
+    - Code: say each character individually (e.g. English: "A B C one two three"; French: "A B C un deux trois")
     ## Normalization rules
-    - "@" symbol → written and spoken as "at"
-    - "." symbol → written and spoken as "dot"
-    - Numbers → written and spoken individually ("123" → "one two three")
-    - Spaces in codes → written and spoken with pauses ("ABC" → "A B C")
+    - "@" symbol → written and spoken as the word for "at" in your language
+    - "." symbol → written and spoken as the word for "dot" in your language
```
gabegma: I don't think this is really accurate, because "at" has a more generic meaning than "@". The word for "at" in French is "à", but in an email, we say "a commercial".
```suggestion
    - "@" symbol → spoken in a linguistically appropriate way for the language, such as "at" in English, "a commercial" in French Canadian, and "arobase" in French.
    - "." symbol → spoken in a linguistically appropriate way for the language, such as "dot" in English and "point" for French.
```

configs/prompts/simulation.yaml
```diff
@@ -275,16 +275,18 @@ user_simulator:
...
+    - Numbers → spoken individually, digit by digit, in a linguistically/culturally appropriate way
+    - Spaces in codes → spoken with pauses between characters
     ## Backup Behavior
-    If the agent mishears you and gets any information wrong, repeat it even slower and spell it out using the NATO phonetic alphabet, one character at a time. ONLY do this if the agent mishears you the first time. 
-    (eg. "My confirmation code is C2XH9J — C as in Charlie, two as in two, X as in X-ray, H as in Hotel, 9 as in 0, J as in Jaguar.")
+    If the agent mishears you and gets any information wrong, repeat it even slower and spell it out using a phonetic alphabet appropriate for your language, one character at a time. ONLY do this if the agent mishears you the first time.
+    For languages that use the Latin alphabet, use the NATO phonetic alphabet (e.g. "C as in Charlie, X as in X-ray").
+    For languages that use non-Latin scripts, spell out characters using their standard phonetic names in your language.
+    (eg. English: "My confirmation code is C2XH9J — C as in Charlie, two, X as in X-ray, H as in Hotel, nine, J as in Juliet.")
```
gabegma: We are repeating the English example both times - I don't think it's meant to be there the second time?

data/itsm_scenarios/10.json
katstankiewicz: I agree it's a pain point but for context, we started translating because some STT services struggle to transcribe a word that's in English because it's expecting French. (eg. `Bonjour, j'ai besoin d'aide pour obtenir un bureau à iCampus.` (when the user said East Campus). For the most part in the runs I think the user uses the English word right now.

katstankiewicz: Later on the user did say campus est (after being prompted) and then the agent translated to English so maybe we don't need the translations: https://snow-core_llm-eva_results_analysis_plusplus.job.console.elementai.com/record_detail?latest_only=false&output_dir=%2Fmnt%2Fvoice_agent%2Fmultilingual&record=20&run=fr-CA_nova-3_gpt-5.4_sonic-3.5%20%C2%B7%20itsm%20%C2%B7%20Clean%20%282026-06-02_22-47-52.765659%29

data/itsm_scenarios/27.json
```diff
@@ -554,7 +577,11 @@
         "slack messenger"
       ],
       "requires_manager_approval": false,
-      "vendor": "Slack"
+      "vendor": "Slack",
+      "name_aliases_translatable": false,
+      "name_aliases_base": [
+        "slack messenger"
```
gabegma: I don't think "messenger" should be here?

data/airline_scenarios/1.2.2.json
raghavm243512: looks like these names were mismatched since the initial commit! The replacement strategy I used was find the name from the user scenario, and use that to fill with placeholders. These ones were different (an error) and so didn't convert. The only edge case (calling about someone else) is in 7.4.1. I have fixed these and handled the edge case

data/itsm_scenarios/10.json
gabegma: Right! This list is not used by the user though right? So it's up to the LLM to either translate or not dynamically? To be more robust, we could have a single mapping from English to other languages for all translated entities (just one word, not many possibilities like right now), and pass it to the user as well.. For example, if we decide "East Campus" is "Campus Est", we add this mapping dynamically to the user prompt when East Campus is detected in the user goal. And we could use the same mapping to avoid hardcoding the name_aliases in the JSON directly, to avoid consistency issues like we have seen. It would also reduce this issue with "iCampus". 

Or we decided that none of the entities should be translated, and we add dynamically to the system prompt the list of entities in a specific user goal that should not be translated.

data/airline_scenarios/1.2.2.json
gabegma: Woah - nice catch!

data/itsm_scenarios/33.json
```diff
@@ -530,10 +549,14 @@
       "department_restrictions": [],
       "name": "Salesforce",
       "name_aliases": [
-        "SFDC"
+        "sfdc"
```
gabegma: I don't think this is right either, although pre-existing issue

configs/prompts/judge.yaml
raghavm243512: I think I missed the mark a little here. The point of fidelity is to see whether the right word was said. The number equivalence isn't the point, so should this "different representation" concept be excluded altogether and focus on allowing only things like "veinteun vs veinteuna", both 21, both much closer in sound, not numerical equivalents that sound super different

data/itsm_scenarios/41.json
```diff
@@ -190,7 +190,7 @@
       "hire_date": "2017-03-13",
       "is_manager": true,
       "last_desk_assignment_date": "2025-01-15",
-      "last_name": "Sj\u00f6berg",
+      "last_name": "Sjöberg",
```
gabegma, line 193: Is this supposed to stay here? Not `<LAST_NAME>`?

raghavm243512: For unrelated employees (not the calling user), things are left as is. Though I suppose it could be odd of Sjoberg is directly reporting to someone in Korea speaking Korean

data/itsm_scenarios/27.json
gabegma: I guess it's fine if we also check the name directly too, and not just the alias?

data/itsm_scenarios/1.json
gabegma: Nevermind - I didn't realize `name_aliases` already existed before this PR

raghavm243512: not sure what you mean here. Both structures look the same (name aliases and name aliases base)

data/itsm_scenarios/27.json
raghavm243512: both are checked

JosephMarinier (approved):
I reviewed the `starting_utterance`s in `fr-CA`, as well as `fr`, for what it's worth.

data/itsm_scenarios/74.json
```diff
@@ -376,7 +376,7 @@
       "hire_date": "2019-06-10",
       "is_manager": false,
       "last_desk_assignment_date": "2026-03-01",
-      "last_name": "De Luca",
+      "last_name": "De <LAST_NAME>",
```
gabegma, line 379: There is an issue here

data/airline_dataset.json
```diff
@@ -584,17 +583,52 @@
...
+    "culture_overrides": {
+      "en": {
+        "first_name": "Samantha",
+        "last_name": "Rodriguez",
+        "phone": "+1-512-555-0148"
+      },
+      "fr-CA": {
+        "first_name": "Maude",
+        "last_name": "Tremblay",
+        "phone": "+1 194 127-7095"
```
gabegma: This can be addressed in a future PR, but it would be good to include valid Quebec area codes for most of them. They are only a few: https://fr.wikipedia.org/wiki/Liste_des_indicatifs_r%C3%A9gionaux_du_Qu%C3%A9bec

Otherwise, it doesn't look like a valid phone number to me.

data/airline_dataset.json
```diff
@@ -16759,17 +17633,52 @@
...
+      "fr": "Bonjour, pouvez-vous me mettre sur le prochain pour rentrer à la maison dès que possible ?"
+    },
+    "culture_overrides": {
+      "en": {
+        "first_name": "Joseph",
+        "last_name": "Scott",
+        "phone": "+1-208-555-5346"
+      },
+      "fr-CA": {
+        "first_name": "Andre",
```
gabegma: I noticed that a lot of the names are missing their accents - this is typically spelled `André`, I wonder if this could confuse the TTS. I also saw `Amelie` instead of `Amélie` earlier, and a few others.

data/airline_dataset.json
```diff
@@ -30194,17 +31838,52 @@
...
+      },
+      "fr-CA": {
+        "first_name": "Stephanie",
+        "last_name": "Rousseau",
+        "phone": "+1 483 643-0099"
+      },
+      "fr": {
+        "first_name": "Valerie",
+        "last_name": "Bernard",
+        "phone": "+33 6 48 36 43 00"
```
gabegma: Same here, it's typically `Valérie` and `Stéphanie` in French.

data/itsm_dataset.json
```diff
@@ -37274,14 +38096,46 @@
         "total_events": 44,
         "total_turns": 21
       }
+    },
+    "starting_utterances": {
+      "en": "Hi, I've got a couple of IT things I need help with.",
+      "fr": "Bonjour, j’ai quelques trucs liées à l’informatique pour lesquelles j’ai besoin d’aide.",
```
gabegma:
```suggestion
      "fr": "Bonjour, j’ai quelques trucs liés à l’informatique pour lesquelles j’ai besoin d’aide.",
```

data/itsm_dataset.json
```diff
@@ -39660,14 +40512,46 @@
         "total_events": 47,
         "total_turns": 23
       }
+    },
+    "starting_utterances": {
+      "en": "Hi, I have a couple of IT things I need help with.",
+      "fr": "Bonjour, j’ai quelques trucs liées à l’informatique pour lesquelles j’ai besoin d’aide.",
```
gabegma:
```suggestion
      "fr": "Bonjour, j’ai quelques trucs liés à l’informatique pour lesquelles j’ai besoin d’aide.",
```

data/itsm_dataset.json
```diff
@@ -58808,7 +60074,7 @@
       },
       "information_required": {
         "first_name": "Marco",
-        "last_name": "De Luca",
+        "last_name": "De <LAST_NAME>",
```
gabegma, line 59781: Same issue in this file - linked to the `De` space issue.

data/itsm_dataset.json
```diff
@@ -73477,14 +75001,46 @@
...
+      "fr-CA": "Bonjour, j’ai quelques demandes TI pour lesquelles j’aurais besoin d’aide."
+    },
+    "culture_overrides": {
+      "en": {
+        "first_name": "Omar",
+        "last_name": "Hassan"
+      },
+      "fr": {
+        "first_name": "Bernard",
+        "last_name": "Andre"
```
gabegma: `André`

scripts/compute_metrics_on_failed.py
```diff
@@ -0,0 +1,158 @@
+"""Compute all registered metrics on a single `_failed_attempt_N` record dir.
```
gabegma, line 1: Was this meant to be included in this PR? It's already really large - adding additional changes like these and the app changes makes the review process more difficult.

src/eva/assistant/pipecat_server.py
```diff
@@ -606,11 +613,6 @@ async def on_audio_data(buffer, audio, sample_rate, num_channels):
             # Accumulate audio chunks
             self._audio_buffer.extend(audio)
             self._audio_sample_rate = sample_rate
-            if len(audio) > 0:
```
gabegma, line 609: Was this removed on purpose?

src/eva/assistant/pipeline/alm_base.py
```diff
@@ -44,6 +48,24 @@
 - If the audio is not clear, respond with exactly: UNCLEAR"""
 
 
+def build_transcription_prompt(language: str | None = None) -> str:
```
gabegma, line 51: Why is this needed in AudioLLM base but not for other pipeline types?

src/eva/user_simulator/client.py
```diff
@@ -184,6 +189,17 @@ async def _run_elevenlabs_conversation(self, api_key: str) -> str:
             else:
                 user_persona = behavior_prompts["default"]
 
+            # Append a language directive to the persona so the simulator speaks
+            # in the target language even if its voice agent could default to English.
+            if self._language and self._language.lower() not in ("en", "english"):
+                try:
+                    lang_display = LANGUAGE_DISPLAY_NAMES[Language(self._language)]
+                except (ValueError, KeyError):
+                    lang_display = self._language
```
gabegma: We have this at least 3 times - could we have a small util for this in the same file as LANGUAGE_DISPLAY_NAMES and import that instead?

tests/unit/utils/test_wer_utils.py
```diff
@@ -113,6 +170,187 @@ def test_10_00_am_not_merged_to_100_am(self):
...
+        result = normalize_text("euh je voudrais un café", Language.FR.value)
+        assert "euh" not in result
+        # diacritics are stripped by the normalizer, so é → e
+        assert "cafe" in result
+
+    def test_abbreviations_expanded(self):
+        assert "madame" in normalize_text("mme dupont", Language.FR.value)
+
+    def test_thousand_separator_dot(self):
+        assert normalize_text("1.000", Language.FR.value) == normalize_text("1000", Language.FR.value)
```
gabegma, line 184: I don't think we use a period for thousand separators, do we? @JosephMarinier

src/eva/assistant/pipeline/alm_base.py
raghavm243512: Cascade pipeline are working with a language code already informed to them, except for intrinsically  multilingual services like an LLM (gpt 5.5 for example). ALMs are similar in that they are natively multilingual (never accept a language code as part of their startup) but they are doing STT in a parallel request in which we inform them of the spoken language

S2S are natively multilingual, but output a transcript on their own (presumably with a system instruction or whatever the implementation chooses to use behind the scene. So this addition is only needed for ALMs

src/eva/models/config.py
```diff
@@ -579,7 +605,8 @@ def _check_companion_services(self) -> "RunConfig":
         # self.model.pipeline_parts is only available if self.model is valid, which the above asserts.
         if "run_id" not in self.model_fields_set:
             suffix = "_".join(v for v in self.model.pipeline_parts.values() if v)
-            self.run_id = f"{datetime.now(UTC):%Y-%m-%d_%H-%M-%S.%f}_{suffix}"
+            lang = self.language.value if self.language != Language.EN else "en"
```
gabegma: Isn't `self.language.value` also working for `Language.EN`?

src/eva/models/config.py
```diff
@@ -579,7 +605,8 @@ def _check_companion_services(self) -> "RunConfig":
...
+            self.run_id = f"{datetime.now(UTC):%Y-%m-%d_%H-%M-%S.%f}_{lang}_{suffix}"
```
gabegma, line 609: Should we not add the language when it's English? I'm just wondering if this change could prevent us from rerunning all of our existing runs.

data/itsm_scenarios/33.json
raghavm243512: The first result when I google those letters, it may be an old fashioned term or something?

data/airline_dataset.json
raghavm243512: I can expand the creation to include a set of realistic area codes if any, and a placeholder too.

data/airline_dataset.json
raghavm243512: This is somewhat by design actually. Not all real data systems take care of characters the same, and people can also just be a little lazy and type the no accent version of their name. So I deliberately do 50% names in pure ascii script, 50% in pure native. (Deterministically seeded which ones we choose from, so same records should always be latin or not)

gabegma: I'm nervous about the TTS side on the user simulator, though. No one would say their name without the accent, that's not realistic.

scripts/compute_metrics_on_failed.py
raghavm243512: I can drop it. whenever I do something useful we don't have I just bring it in. Probably not ideal when feature is this large though

src/eva/models/config.py
raghavm243512: rerunning existing would require us providing the run_id, so the suffix only affects future runs

scripts/compute_metrics_on_failed.py
gabegma: You can keep it for this PR - but I'm not reviewing any of those files, given the importance of the rest of the changes. Although it's easy for the author to add all changes to the same PR, it transfers the burden to the reviewer (and ultimately to the author as well, who ultimately can't merge as fast as he needs to address many more comments). And when PRs are large, this is typically where we introduce errors. I always prefer a lot more smaller PRs :)

src/eva/models/config.py
gabegma: I see, perfect!

src/eva/assistant/pipeline/alm_base.py
gabegma: Ah I see - although as a nuance, the AudioLLM itself is not using/needing that STT - it's for our logs mostly. Thanks!

data/itsm_scenarios/33.json
gabegma: Ok cool! Now that I know we are also using the name too, it's fine :)

src/eva/assistant/pipeline/alm_base.py
raghavm243512: We do use those transcripts for future turns I believe though, that's the only way the agent sees what happened before the current audio

src/eva/assistant/pipecat_server.py
raghavm243512: fixed

tests/unit/utils/test_wer_utils.py
raghavm243512: Somewhat mixed opinions here: https://www.reddit.com/r/French/comments/114sir7/decimal_points_in_numbers/?rdt=38020

data/airline_dataset.json
raghavm243512: I can make this work so that only on the database side will the names sometimes have/not have the accent. User simulator always sees it's native name

tests/unit/utils/test_wer_utils.py
gabegma: It does mention that's not in Quebec, referring to this official link: https://vitrinelinguistique.oqlf.gouv.qc.ca/21191/la-typographie/nombres/emploi-du-signe-decimal-et-indication-de-la-division-dans-les-unites-de-mesure

Is this applied to both French and French Canadian? It's a detail, but I would put space to avoid any issues.

src/eva/assistant/pipeline/alm_base.py
gabegma: Right - thanks for reminding me!

data/itsm_dataset.json
raghavm243512: This is intended to be for all occurrences right? Or was there something specific here that makes it different?

tests/unit/utils/test_wer_utils.py
raghavm243512: Language.FR is French French (EU). They have separate rules but both of them show this:
```
"decimal_separator": "comma",
  "thousand_separators": [
    "dot",
    "space"
  ]
```

raghavm243512:
@gabegma new design to fix accent issue for user while keeping the real world potential of lacking accents on database side:

romanized placeholders are stored in DB rather than done at generation time. More deterministic, readable, cleaner, etc. and user is always seeing native scrip

data/itsm_aliases/a_garage.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "A Garage",
```
gabegma: Thanks for doing this - I'm still thinking we should pass the translation to the user too, because otherwise they might not match, but we can tackle it in a future PR

data/itsm_aliases/garage_a.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "Garage A",
```
gabegma, line 2: We have A Garage and Garage A?

configs/prompts/simulation.yaml
```diff
@@ -375,16 +376,18 @@ user_simulator:
...
```
gabegma: This is still outdated - and there is another one below.

configs/prompts/simulation.yaml
```diff
@@ -375,16 +376,18 @@ user_simulator:
...
-    - Code: "A B C one two three"
+    - Phone: say each digit individually in a linguistically/culturally appropriate way (e.g. English: "five five five... one two three... four five six seven"; Korean: "공 일 공... 일 이 삼... 사 오 육 칠")
+    - Code: say each character individually (e.g. English: "A B C one two three"; French: "A B C un deux trois")
     ## Normalization rules
-    - "@" symbol → written and spoken as "at"
-    - "." symbol → written and spoken as "dot"
-    - Numbers → written and spoken individually ("123" → "one two three")
-    - Spaces in codes → written and spoken with pauses ("ABC" → "A B C")
+    - "@" symbol → written and spoken as the word for "at" in your language
+    - "." symbol → written and spoken as the word for "dot" in your language
```
gabegma: Same for this

data/itsm_aliases/garage_a.json
raghavm243512: And apparently an "Alpha garage" as well, all from original dataset. I will definitely remove the A Garage, should we do same for Alpha? It is "different" but not really

data/airline_scenarios/1.1.2.json
```diff
@@ -30,12 +30,12 @@
       "passengers": [
         {
           "elite_status": null,
-          "email": "samantha.rodriguez@example.com",
-          "first_name": "Samantha",
-          "last_name": "Rodriguez",
+          "email": "<FIRST_NAME_ROMANIZED>.<LAST_NAME_ROMANIZED>@example.com",
+          "first_name": "<FIRST_NAME_ROMANIZED>",
+          "last_name": "<LAST_NAME_ROMANIZED>",
```
gabegma, line 35: Why do we have romanized here now?

data/itsm_dataset.json
gabegma: No, not for all - it's because "truc" is masculine here, so we need to drop the `e`.

tests/unit/utils/test_wer_utils.py
gabegma: Ok well at least remove dot for FR CA.

data/airline_scenarios/1.1.2.json
raghavm243512: Real world data may not always keep native script. I can say my name is राघव and the agent should know I might mean "Raghav". So I made some of it romanized in DB and some not

gabegma: Oh I didn't see your comment about the new design - so that's on purpose? It doesn't seem like it was applied to all files.

gabegma: Won't this be an issue when computing the hash for the expected DB state or for authentication success if we can have either?

data/itsm_aliases/garage_a.json
gabegma: I would keep alpha as different

data/itsm_aliases/intellij_idea.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "IntelliJ IDEA",
+  "translatable": false,
+  "base": [],
```
gabegma: This one doesn't have a base? Also, it's a bit annoying to have IntelliJ, IntelliJ IDEA and JetBrains IntelliJ IDEA.

data/airline_scenarios/1.1.2.json
raghavm243512: no I tested the task completion in French with a romanized expectation. Even with the accent (on the letter) task completion passed. The agent is told to try the romanized name spelling if needed. I didn't do it to all files because both could be valid depending on the database behind the scenes

data/itsm_aliases/intellij_idea.json
raghavm243512: yea the aliases are actually all pretty messed up. Should I redo them a bit more thoroughly? We have things like 'Downtown Garage A', 'Garage A', 'A Garage', guess we never really noticed the data before since it was all separated and duplicated

gabegma: I'm still confused why we need alias in the first place, I wonder if we could change the tools all together, and also avoid the translation issues. I'm ok leaving for a next PR but I would tackle this right up next.

raghavm243512: Claude suggested consolidation (which I agree with, lmk if any concerns):

IntelliJ trio → keep intellij_idea.json ("IntelliJ IDEA"). Delete intellij.json and jetbrains_intellij_idea.json. All three refer to the same JetBrains product; "IntelliJ IDEA" is the canonical product name.

main_garage_a.json → garage_a.json ("Garage A"). main_garage_a has base ['garage a', 'main garage'] and garage_a has base ['a garage', 'main garage'] — same aliases, just calling it "Main Garage A" vs "Garage A". Identical to the a_garage/garage_a case we just fixed.

east_garage.json → east_campus_garage.json ("East Campus Garage"). Both share east garage as a base alias; east_garage has no additional distinguishing info.

north_lot.json → north_surface_lot.json ("North Surface Lot"). north_surface_lot has north lot as a base alias, and north_lot's alias north lot at headquarters suggests they're the same spot.


Or at least I can make the separation a bit clearer for main garage A and garage A for example

data/airline_scenarios/1.1.2.json
gabegma: Is it possible to add a test that checks for a disconnect between the JSON individual files and the actual dataset? I think it would be hard to spot right now if one of them as romanized but not the other one?

data/itsm_aliases/intellij_idea.json
gabegma: yep agreed

data/itsm_aliases/microsoft_visio.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "Microsoft Visio",
+  "translatable": false,
+  "base": [],
```
gabegma: Some of those lost their base - this used to have `visio` here

data/itsm_aliases/adobe_creative_cloud.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "Adobe Creative Cloud",
+  "translatable": false,
+  "base": [],
```
gabegma: This also used to have aliases `creative cloud` and `adobe cc`

data/airline_scenarios/1.1.2.json
raghavm243512: placeholders will be static and unchanging but I can add a general thing that takes the initial DB, applies expected trace, and should pass per language? or something along these lines

src/eva/utils/culture.py
```diff
@@ -0,0 +1,293 @@
+"""Culture/language placeholder resolution for evaluation records.
```
gabegma, line 1: It would be great to add a few tests for these functions - could be included in the other test we just talked about, or not.

src/eva/__init__.py
gabegma: We should bump this one too

gabegma (approved):
Excellent work!! I love that you have pushed your script so others can add new languages.

I'm still nervous about the aliases, and I think we should feed the translations to the user so they match. Or we change the tools themselves so we don't need to pass unstructured text.
I would do that before adding new languages, but in a separate PR.

I also think we are missing a few tests - especially for looking at the disconnect between the dataset's individual files and the main one, but I need to drop, so I'm pre-approving!

PR #130, opened 2026-05-29
Add confidence intervals to main code

lindsaydbrin:
Every `eva metrics` run now emits 95% bootstrap confidence intervals for composite and main metrics (not sub-metrics) in `metrics_summary.json`. CIs are deterministic within a run (same `run_dir.name` → byte-identical bounds) and independent across runs.

**Scope** (purely additive — no existing field renamed or changed):
- `overall_scores.{COMPOSITE}.{mean_ci_lower, mean_ci_upper, mean_ci_n_scenarios}` - for 6 composites: [`EVA-A`, `EVA-X`, `EVA-overall`] x [`_pass`, `_mean`]
- `overall_scores.pass_k.{COMPOSITE}.{stat}_ci_{lower,upper}` - for `pass_at_1`, `pass_at_k`, `pass_power_k_observed` (multi-trial only). `pass_power_k_theoretical` does not have CIs (clearly).
- `per_metric.{name}.{mean_ci_lower, mean_ci_upper, mean_ci_n_scenarios}` for all metrics (typically/currently: `task_completion`, `faithfulness`, `agent_speech_fidelity`, `conversation_progression`, `turn_taking`, `conciseness`).
- `per_metric.{name}.pass_k.{stat}_ci_{lower,upper}` (k>1 only, again of course).

**Statistical design** (matches the approach on `analysis/eva-bench-stats/` / in the paper): percentile bootstrap on per-scenario sample means, `N_BOOT=2000`, `ALPHA=0.05`, seed derived from `run_dir.name` via SHA-256 (cross-process-stable).


src/eva/__init__.py
```diff
@@ -11,4 +11,4 @@
 
 # Bump metrics_version when changes affect metric computation (metrics code,
 # judge prompts, pricing tables, postprocessor).
-metrics_version = "2.0.0"
+metrics_version = "2.1.0"
```
lindsaydbrin: Is this correct since it changes output fields?

lindsaydbrin:
@tara-servicenow asked about handling missing/errored samples. Here's Claude's explanation: 

> How missing/errored trials are handled in CI computation
> CI computation uses get_score() to filter raw trial data: errored trials (score.error set) and missing values return None and are dropped. Valid trials are grouped by scenario ID and averaged to produce one value per scenario; scenarios where every trial is None are dropped entirely. The bootstrap CI is then computed over those per-scenario means.
>
> This matches what the main metric aggregation does — both exclude errors and missing values before computing statistics — with two differences worth noting:
>
> 1. Granularity. The main loop counts error_count and missing_count separately and tracks totals in the output. The CI path collapses both to None with no separate tracking; that distinction is only visible in the main aggregation stats.
> 2. Unit of analysis. The main loop treats each record/trial as an independent data point for the mean. The CI groups trials by scenario first and averages them, so each scenario contributes one value regardless of how many trials it has. This is the statistically correct unit for the bootstrap — a design choice, not a discrepancy.


You can see the `get_score()` on line 116 of `aggregation.py`:
```
    return _scenario_means(
        {record_id: record_metrics.get_score(metric_name) for record_id, record_metrics in all_metrics.items()}
    )
```

and then lines 96-97 in `_scenario_means()`:
```
        if val is None:
            continue
```
@gabegma does this make sense to you? @tara-servicenow mentioned that you had thought about handling `None` vs. errors.

tara-bogavelli:
High level this looks good to me but will let Gab review more carefully

src/eva/utils/bootstrap.py
```diff
@@ -0,0 +1,99 @@
...
+    """95% bootstrap CI on the mean (default alpha=0.05).
+
+    ``seed`` is keyword-only and required: callers must supply a deliberate
+    seed (typically from ``run_seed(run_dir.name)``) so behavior is deterministic.
+
+    Returns ``(lower, upper)``; ``(nan, nan)`` if the input is empty.
+    """
+    boot = bootstrap_resample(values, n_boot=n_boot, seed=seed)
+    if len(boot) == 0:
+        return float("nan"), float("nan")
```
gabegma: Any reason for returning `nan` rather `None` here? `NaN` is not valid JSON so this could cause an issue if it reaches it.

gabegma: The rounding would break after, but you could round directly here on row 63 instead.

src/eva/utils/bootstrap.py
```diff
@@ -0,0 +1,99 @@
...
+    """Return ``{name}_ci_lower`` / ``{name}_ci_upper`` for each ``(name, sample)`` pair."""
+    out: dict[str, float] = {}
+    for name, sample in samples.items():
+        lower, upper = bootstrap_ci(sample, seed=seed)
+        out[f"{name}_ci_lower"] = round(lower, decimals)
+        out[f"{name}_ci_upper"] = round(upper, decimals)
+    return out
+
+
+def mean_ci_fields(
```
gabegma, line 70: Small suggestion - perhaps this function should be with the other functions in aggregation.py? The back and forth makes it harder to read, and this deals with `scenario_values` that come directly from there.

gabegma: Actually, never mind, I think it's fine.

src/eva/metrics/aggregation.py
```diff
@@ -83,6 +86,55 @@ class EVACompositeDefinition:
 ]
 
 
+def _scenario_means(per_record_values: dict[str, float | None]) -> np.ndarray:
```
gabegma: I was a bit nervous with these functions returning numpy arrays here, as it required a bit of digging to make sure those are correctly handled and not kept as numpy arrays. Small suggestion to use list[float] for the functions in this file, and make the CI functions accept `Sequence[float]`.

gabegma (approved):
LGTM! Only tiny comments

src/eva/metrics/aggregation.py
lindsaydbrin: Sounds good! On it.

src/eva/utils/bootstrap.py
lindsaydbrin: I took your suggestion to return `None` directly, but rather than dealing with it in the rounding, the code now explicitly checks for `None` values where they're possible (more parallel to how it's handled in `mean_ci_fields`). See lines 60-63 in `bootstrap.py`.

I also:
* Renamed `bootstrap_ci_fields` to `named_ci_fields` so that it's clear how it's differentiated from `mean_ci_fields` (it's all bootstrap so that meant nothing)
* Cleaned up docstrings where they were basically just redundantly describing the code

Does this look reasonable to you?

gabegma: Yes - looks great!

gabegma:
I think your dropped your metric version bump while resolving conflicts with main

PR #134, opened 2026-06-04
speed up script for regen metric signatures

katstankiewicz:
Script itself was quick but imported pipecat and all of eva which was slow ~4-5s.
This removes some unused imports and add stubs
Brings down the time to ~1-2s

### DIFF
--- a/scripts/regen_metric_signatures.py
+++ b/scripts/regen_metric_signatures.py
@@ -11,9 +11,81 @@ Usage:
 """
 
 import json
+import sys
+from importlib.abc import Loader, MetaPathFinder
+from importlib.machinery import ModuleSpec
 from pathlib import Path
+from types import ModuleType
+from typing import TypedDict
 
-from eva.metrics.signatures import compute_all_metric_signatures
+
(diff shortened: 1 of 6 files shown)


scripts/regen_metric_signatures.py
```diff
@@ -11,9 +11,81 @@
 """
 
 import json
+import sys
+from importlib.abc import Loader, MetaPathFinder
+from importlib.machinery import ModuleSpec
 from pathlib import Path
+from types import ModuleType
+from typing import TypedDict
```
weiz9: I tested the PR on python 3.11 and crashed with PydanticUserError: Please use typing_extensions.TypedDict. Do we want to use "typing_extensions import TypedDict" instead since pyproject.toml supports 3.11 and DeploymentTypedDict is used as a pydantic base. Other than that everything looks good, went from 8s -  3s locally.

katstankiewicz: thank you! I updated to use `typing_extensions import TypedDict`

weiz9 (approved):
Looks good, thank you!

PR #135, opened 2026-06-04
Evaluate timeout records

katstankiewicz:
- Set a max number of attempts where a conversation reaches conversation timeout
- Set a default of 1 because this is a model error and we want model stats based on this run, ie task_completion, conciseness etc. 
- Sets a hard cap for conversation length for models to adhere to
- User behaviour and speech fidelity must still be 1.0 to get evaluated
  - If  `max_timeout_attempts` is > 1, keep a cache of any previous runs that had timeout reached and user_behaviour == 1.0 so we can evaluate that run instead of trying again after meeting max_timeout_attempts
    - eg. max_timeout_attempts == 2, first run timesout and User behaviour == 1.0 but second run also timesout and user behaviour is 0. In this case don't run again, take the first run
- Add `conversation_finished_on_time` as a diagnostic metric to see how many finished on time
- Save `timeout_accepted` in result.json to show when a conversation timedout but we evaluate it
- Update user behaviour fidelity metric to handle four conversation end modes: error, timeout, inactivity timeout and user ended call
- Update default conversation timeout to 10 minutes (600s)


raghavm243512 (approved):
lgtm, small word choice though: I think framing it as a "duration" vs a time out would be more clear. Timeout usually means "waiting for something and not getting it" at least to me.

gabegma:
I agree that "time-out" doesn't seem like the right terminology

katstankiewicz:
I updated to use "time_limit" instead of "timeout". It definitely makes more sense for the metric though 'timeout' comes from the asyncio call that times out waiting for the call to end ie:
```
conversation_ended_reason = await asyncio.wait_for(
     self._run_conversation(),
     timeout=self.config.conversation_time_limit_seconds,
)
```

PR #136, opened 2026-06-04
Migrate data/*_dataset.jsonl to data/*_dataset.json

JosephMarinier:
Convert the evaluation datasets from JSONL (one object per line) to pretty-printed JSON arrays, and update the loader/path accordingly. Content is unchanged.

I cherry-picked these changes from #121 to make it easier to then review that PR. The only addition is adapting tests/unit/test_current_date_consistency.py, which was silently skipping all tests because of this, which I now removed:
```py
        if not dataset_path.exists():
            continue
```
Here is how you can review these files's changes in #121:
```sh
git diff joseph/migrate-dataset-jsonl-to-json pr/multilingual -- data/*_dataset.json
```


gabegma (approved):
Thank you!!!

PR #137, opened 2026-06-08
Fix speech fidelity intended text with newlines

JosephMarinier:
In the speech fidelity judge prompt, turns are separated by newlines, so a turn whose text contains its own newline would spill onto unlabeled lines and, sometimes, be misread by the judge, unfairly penalizing speech fidelity.

---
For example:
```
Turn 1: Hello,

World!
Turn 2: Nice to meet you, too.
```
The judge could get confused and only expect "Hello".

---
As a concrete example, we noticed some LLMs repeating their output after a newline:
```
Turn 1: Absolutely. I can help with access requests. First, I need to verify your identity with your employee ID and the last four digits of your phone number on file.
Absolutely. I can help with access requests. First, I need to verify your identity with your employee ID and the last four digits of your phone number on file.
```
And the speech fidelity judge made it seem like it was a TTS issue, whereas it was an LLM issue:
> The assistant repeats the entire response twice in the audio, which is a significant deviation from the intended text.

---
Flattening the text to a single line avoids this confusion, and the speech fidelity judge considers the whole text, as expected. Not that, in the example above, this turn is penalized for conciseness, as expected.

### DIFF
--- a/src/eva/metrics/speech_fidelity_base.py
+++ b/src/eva/metrics/speech_fidelity_base.py
@@ -387,5 +387,15 @@ class SpeechFidelityBaseMetric(AudioJudgeMetric):
 
     @staticmethod
     def _format_intended_turns(intended_turns: dict[int, str]) -> str:
-        """Format intended turns dictionary as numbered list."""
-        return "\n".join(f"Turn {turn_id}: {text}" for turn_id, text in intended_turns.items())
+        """Format intended turns dictionary as a numbered list, one turn per line.
+
+        Each turn's text is flattened to a single line (internal newlines and
+        repeated whitespace collapsed to single spaces). Turns are separated by
+        newlines, so a turn whose text contains its own newline -- e.g. a normal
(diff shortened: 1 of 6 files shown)


fanny-riols (approved):

src/eva/metrics/speech_fidelity_base.py
```diff
@@ -387,5 +387,15 @@ def _get_intended_turns(self, context: MetricContext) -> dict[int, str]:
 
     @staticmethod
     def _format_intended_turns(intended_turns: dict[int, str]) -> str:
-        """Format intended turns dictionary as numbered list."""
-        return "\n".join(f"Turn {turn_id}: {text}" for turn_id, text in intended_turns.items())
+        """Format intended turns dictionary as a numbered list, one turn per line.
```
gabegma, line 390: LGTM! Let's bump the version of the speech fidelity metric, too, especially if you've seen issues for some models. Especially because the code hash and prompt template hash are the same.

gabegma (approved):

PR #138, opened 2026-06-08
Add optional OpenAI Realtime user simulator

PASFIELD-OAI:
## Summary

This PR adds OpenAI Realtime as an optional provider for EVA's simulated caller while preserving ElevenLabs as the default.

- adds a configurable `openai_realtime` user simulator provider
- keeps the existing ElevenLabs caller and historical `UserSimulator` import compatible
- introduces a provider-neutral caller interface and `user_simulator_events.jsonl` artifact
- continues writing `elevenlabs_events.jsonl` for ElevenLabs runs
- updates validation and metric consumers to accept the neutral artifact with legacy fallback
- adds lifecycle, timeout, response-sequencing, audio, artifact-parity, and configuration tests
- documents configuration, supported perturbations, and a one-record smoke command

## Motivation

EVA currently requires ElevenLabs for user simulation. This gives benchmark users an OpenAI Realtime alternative without changing the assistant implementation, judge configuration, or metric definitions.

## Compatibility

- ElevenLabs remains the default provider.
- Existing ElevenLabs environment variables and artifacts remain supported.
- Assistant frameworks, Pipecat services, and judges are unchanged.
- Behavior, background-noise, and connection-degradation perturbations work with the OpenAI caller. Accent variants remain ElevenLabs-only because they depend on dedicated ElevenLabs agents.

## Validation

- `uv sync --frozen --extra dev`
- `pytest -q -ra`: 1,636 passed, 52 skipped, 3 expected failures
- `ruff check .`
- `ruff format --check .`
- `pre-commit run --all-files`
- `git diff --check`
- one live OpenAI Realtime caller smoke completed with `conversation_ended_reason: "goodbye"`, `conversation_valid_end = 1.0`, and no caller error

The 52 skipped tests are existing airline schema checks whose fixture path resolves to an empty parameter set. The three expected failures are existing documented WER-normalization limitations.

## Review note

I did not have ElevenLabs credentials available for a live default-provider regression. The ElevenLabs compatibility path is covered by unit and artifact-parity tests, but a maintainer-side one-record ElevenLabs smoke would close the remaining runtime validation gap as I was unable to test


PASFIELD-OAI:
@fanny-riols, Per our convo last week - here's the PR for the OpenAI Realtime option (rather than ElevenLabs) - please let me know if you have any feedback! See note in the PR about my lack of ElevenLabs access for testing

.env.example
```diff
@@ -234,6 +234,24 @@ EVA_MODEL__LLM=gpt-5.2
...
+#d string
+EVA_USER_SIMULATOR__MODEL=gpt-realtime-1.5
+
+#i OpenAI voice for the female caller persona.
+#d string
+EVA_USER_SIMULATOR__FEMALE_VOICE=marin
+
+#i OpenAI voice for the male caller persona.
+#d string
+EVA_USER_SIMULATOR__MALE_VOICE=cedar
```
raghavm243512:
```suggestion
# --- User simulator provider ---
#i Caller provider. ElevenLabs is the backward-compatible default.
#d enum
#e elevenlabs,openai_realtime
EVA_USER_SIMULATOR__PROVIDER=elevenlabs

#i OpenAI Realtime caller model. Used only when provider=openai_realtime.
#d string
#x EVA_USER_SIMULATOR__PROVIDER=openai_realtime
#v EVA_USER_SIMULATOR__MODEL=gpt-realtime-1.5

#i OpenAI voice for the female caller persona.
#d string
#x EVA_USER_SIMULATOR__PROVIDER=openai_realtime
#v EVA_USER_SIMULATOR__FEMALE_VOICE=marin

#i OpenAI voice for the male caller persona.
#d string
#x EVA_USER_SIMULATOR__PROVIDER=openai_realtime
#v EVA_USER_SIMULATOR__MALE_VOICE=cedar
```

gabegma:
Thank you Wesley, great PR!! I tested it and it seems to be working well on a few samples.
  
We'll push a few commits to resolve some conflicts with changes on main. We'll also take the opportunity to stop producing `elevenlabs_events.jsonl` for new runs (to avoid duplication), while keeping support for loading it in older runs.

Before officially recommending this as an alternative user simulator, we'd want to analyze a larger number of conversations to rule out new failure modes (see below). That said, we're happy to merge this so it's available for those who want to try it!

**New Failure Modes**
There might be other implications I'm not thinking about at the moment, but here is what comes to mind for now:

_Validation metrics_
* The existing validation metrics were built for cascade pipelines and might need updates to work with S2S — we can address this in a follow-up PR. 
* It's worth noting that UserBehavioralFidelity was only validated on existing agents, so it may not catch failure modes specific to GPT-Realtime. GPT-Realtime may struggle with instruction-following (we've seen this when benchmarking it as the agent), though the user scenarios may be simple enough that this isn't an issue in practice — and if the validation metrics do catch it, those conversations would just be retried.
* We are now working with a "transcript" of what the user said, rather than what was intended (which we get in Cascade), so UserBehavioralFidelity could mistakenly flag transcription issues as behavioral issues.

_Regular metrics_
* Merging the user simulator logs with the agent logs to build the conversation trace is delicate, and we've hit issues in the past with delayed or missing transcriptions. We'll need to verify this new setup doesn't introduce new problems there. 
* Assistant turn transcripts are now produced by OpenAI's input audio transcription (Whisper), rather than ElevenLabs ScribeV2.2Realtime. If this introduces more transcription errors on assistant turns, it could unfairly penalize the agent on text-based metrics, although our prompts try to guard against it. 
* Given we no longer have access to the intended text for the user turns, transcription issues in the user turns could cause issues for some metrics like faithfulness and conversation progression. For example, if the speech says "DJ3LPO" but the GPT-realtime transcription service transcript says "DJLPO", then the agent who tries "DJ3LPO" in a tool call would be unjustifiably marked as unfaithful. Our speech fidelity validation metric could potentially detect this, but audio judges are not always the most reliable.

PASFIELD-OAI:
"Before officially recommending this as an alternative user simulator, we'd want to analyze a larger number of conversations to rule out new failure modes (see below). That said, we're happy to merge this so it's available for those who want to try it!"

This makes total sense to me! Please let me know if you need anything else from me, but it seems like getting people to test with different setups that are hard to reproduce locally would be a great next step!

raghavm243512 (approved):

gabegma (approved):
Thank you again for this awesome addition!! Before merging, I'll let @katstankiewicz take a last look at the user simulator section, but the rest is good to go on my end!

src/eva/orchestrator/worker.py
```diff
@@ -12,14 +12,16 @@
...
-from eva.user_simulator.client import UserSimulator
+from eva.user_simulator.factory import create_user_simulator
 from eva.utils.culture import resolve_scenario_db, resolve_user_config, resolve_user_goal
 from eva.utils.error_handler import create_error_details
 from eva.utils.hash_utils import get_dict_hash
 from eva.utils.logging import add_record_log_file, current_record_id, get_logger, remove_record_log_file
 
 logger = get_logger(__name__)
 
+USER_SIMULATOR_SHUTDOWN_GRACE_SECONDS = 90
```
katstankiewicz:
```suggestion
USER_SIMULATOR_SHUTDOWN_GRACE_SECONDS = 20
```
90 seconds seems like a lot of time to shut down. Maybe 20 seconds makes more sense?

katstankiewicz (approved):
did some minor clean up but looks good!

PR #139, opened 2026-06-08
Add support for Cartesia Ink-2 (semantic endpointing) STT + 16 kHz input-rate fix

sauhardjain:
## Changes

This PR updates the Cartesia STT path used by the cascade benchmark. `EVA_MODEL__STT=cartesia` now uses Cartesia ink-2, while the previous ink-whisper path remains available as `EVA_MODEL__STT=cartesia-multilingual`.

It also fixes the sample rate declared to Cartesia STT services. The pipeline feeds 16 kHz input audio, so Cartesia STT now declares 16 kHz by default while still allowing an explicit `sample_rate` override in `EVA_MODEL__STT_PARAMS`.

The ink-2 path is self-endpointing, so selecting `cartesia` automatically uses external turn start and stop strategies and disables local VAD. The Pipecat server logs ink-2 turn diagnostics, but those diagnostics do not change aggregation behavior.

## Testing

A useful validation is to run a small cascade evaluation with `EVA_MODEL__STT=cartesia` and `EVA_MODEL__STT_PARAMS='{"api_key":"...","model":"ink-2"}'`, then confirm that the resolved config uses external turn start and stop strategies with `VAD=none`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -64,14 +64,14 @@ AWS_SECRET_ACCESS_KEY=your_aws_secret_access_key_here
 # --- LLM mode: STT ---
 #i STT provider for the voice pipeline.
 #d enum
-#e assemblyai,cartesia,deepgram,deepgram-flux,elevenlabs,nvidia,nvidia-baseten,openai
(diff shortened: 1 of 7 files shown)


sauhardjain:
cc @fanny-riols as discussed last week!

PR #140, opened 2026-06-08
Add cascade latency levers and streaming robustness

sauhardjain:
## Summary

This PR adds opt-in latency controls for cascade (STT -> LLM -> TTS) systems and fixes audit/history edge cases that can distort benchmark results. The default cascade configuration remains unchanged unless these flags are enabled.

## Motivation

The cascade harness can add avoidable latency when it waits for a full LLM response before sending text to TTS, or when tool calls leave the caller in silence. It also does not expose the parallel tool-call setting used by the ElevenAgents cascade assistant configuration, which makes cross-harness comparisons harder to interpret.

Interrupted and tool-heavy turns need careful handling as well. An interrupted transfer can leave an assistant tool call without a matching tool result, and streamed speech can reach TTS before the LLM call is cancelled or fails. This PR keeps the next model request valid and keeps the audit log aligned with emitted audio.

## Changes

- Adds `EVA_MODEL__PRE_TOOL_SPEECH` to let the model produce a brief lead-in before tool calls. The lead-in is model-generated, not templated filler.
- Adds `EVA_MODEL__LLM_STREAMING` for sentence-level Chat Completions streaming to TTS. Responses API deployments warn and use the existing non-streaming path.
- Adds `EVA_MODEL__PARALLEL_TOOL_CALLS` to forward the provider setting when tools are present. Leaving it unset preserves provider defaults; setting it to `false` matches the ElevenAgents assistant config.
- Repairs orphaned assistant tool calls before replaying history to the model, while ignoring malformed non-string tool result IDs.
- Records any streamed text that was already emitted if a stream is cancelled or fails mid-turn.
- Updates transcript processing so fully spoken streamed responses can span multiple TTS segments.
- Documents the new flags in `.env.example` and the experiment setup guide.

## Reviewing

Run one cascade record with the new flags unset and confirm that default behavior is unchanged.

Then run a tool-using cascade record with `EVA_MODEL__PRE_TOOL_SPEECH=auto`, `EVA_MODEL__LLM_STREAMING=true`, and `EVA_MODEL__PARALLEL_TOOL_CALLS=false`. Check that speech can begin before the final LLM response is complete, that the audit log records the assistant response, and that interrupted streamed speech is preserved instead of replaced by a generic error.


sauhardjain:
cc @fanny-riols as discussed last week!

src/eva/assistant/pipeline/services.py
```diff
@@ -128,21 +124,32 @@ def create_stt_service(
...
+        logger.info(f"Using Cartesia STT (ink-2): {model_name}")
+        return CartesiaTurnsSTTService(
+            api_key=api_key,
+            sample_rate=params.get("sample_rate", 16000),
+            should_interrupt=params.get("should_interrupt", True),
+            settings=CartesiaTurnsSTTService.Settings(model=model_name),
+        )
+
+    elif model_lower == "cartesia-multilingual":
+        logger.info(f"Using Cartesia multilingual STT (ink-whisper): {params['model']}")
```
```suggestion
        logger.info(f"Using Cartesia multilingual STT: {params['model']}")
```
we are already displaying the model so we don't need to display twice

src/eva/assistant/pipeline/services.py
```diff
@@ -128,21 +124,32 @@ def create_stt_service(
...
+                sample_rate=SAMPLE_RATE,
+                speech_model=params["model"],
             ),
         )
 
     elif model_lower == "cartesia":
-        logger.info(f"Using Cartesia STT: {params['model']}")
+        # ink-2 provides its own turn boundaries; ModelConfig selects external endpointing.
+        model_name = params.get("model", "ink-2")
+        logger.info(f"Using Cartesia STT (ink-2): {model_name}")
```
```suggestion
        model_name = params["model"]
        logger.info(f"Using Cartesia STT: {model_name}")
```
model is a required param so we can just use that value directly

src/eva/assistant/agentic/system.py
```diff
@@ -92,9 +143,23 @@ def __init__(
             agent_instructions=agent.instructions,
             datetime=self.current_date_time,
         )
+        if self.pre_tool_speech == "auto":
+            self.system_prompt += (
+                "\n\n## Responsiveness\n"
+                "Before calling a tool, say one brief sentence first. For lookups, use a short lead-in "
+                '(e.g. "Let me pull that up."). Before any action that changes a booking or charges money, '
+                "state what you'll do and the cost, get confirmation, then call the tool."
```
katstankiewicz: We can add this to the prompt manager in configs/prompts/simulation.yaml under agent.pre_tool_speech

katstankiewicz (approved):

PR #141, opened 2026-06-09
Fix time exceeded for re-runs

katstankiewicz:
- Allow time-exceeded runs to skip the conversation valid end gate when re-running  (this happens in the `validate_existing` flow)
- Save _conversation_stats immediately if timeout is hit (otherwise they were being saved as 0)

### DIFF
--- a/src/eva/orchestrator/validation_runner.py
+++ b/src/eva/orchestrator/validation_runner.py
@@ -1,6 +1,7 @@
 """Validation metrics runner for benchmark validation mode."""
 
 import asyncio
+import json
 from dataclasses import dataclass, field
 from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


src/eva/orchestrator/validation_runner.py
```diff
@@ -72,28 +73,50 @@ async def run_validation(self) -> dict[str, ValidationResult]:
...
+        for record_id in not_finished:
+            result_path = self.run_dir / "records" / record_id / "result.json"
+            try:
+                with open(result_path) as f:
+                    result_data = json.load(f)
+                if result_data.get("conversation_ended_reason") == "time_limit_exceeded":
+                    time_limit_ids.append(record_id)
+                    continue
+            except (FileNotFoundError, json.JSONDecodeError):
+                pass
```
gabegma: Detail - but I checked in other places in the code  - and we don't do try/except when loading the result file, it seems to always be valid json.
```suggestion
            if result_path.exists():
                with open(result_path) as f:
                    result_data = json.load(f)
                if result_data.get("conversation_ended_reason") == "time_limit_exceeded":
                    time_limit_ids.append(record_id)
                    continue
```

src/eva/orchestrator/validation_runner.py
```diff
@@ -72,28 +73,50 @@ async def run_validation(self) -> dict[str, ValidationResult]:
...
         )
 
-        for record_id in not_finished:
+        for record_id in truly_not_finished:
             validation_results[record_id] = ValidationResult(passed=False)
 
-        if gate_passed:
+        # Run LLM metrics on gate-passed and time-limit records together.
+        # gate_passed records get GATE_METRIC=1.0 in their scores; time-limit records do not.
+        all_llm_ids = gate_passed + time_limit_ids
```
gabegma: Perhaps a name like `ids_to_judge` would be clearer? I was confused with what `all_llm_ids` meant at first. Same for `all_llm_contexts`

gabegma (approved):
LGTM!!

PR #142, opened 2026-06-09
Small adjustments for multilang services

raghavm243512:
Using the enum instead of the string allows pipecat to auto resolve some of the name formats for languages.

Added support for Deepgram Flux Language param

Updated Dockerfile for better iteration (push size goes down to 50mb from 1.2gb when we change the local source)

Update pydantic config to support our airline traveling companion. (One record, a woman calls about her husband, and he needs cultural naming as well from the earlier multilingual PR)

Remove default user and instead use en user

update readme to be more accurate 

Fix ALM parameter for thinking hard coded to False

### DIFF
--- a/src/eva/assistant/pipeline/services.py
+++ b/src/eva/assistant/pipeline/services.py
@@ -176,7 +176,10 @@ def create_stt_service(
             flux_settings_kwargs: dict[str, Any] = {"model": params["model"]}
             # Flux ignores `language`; only `flux-general-multi` honors `language_hints`.
             if params["model"] == "flux-general-multi":
-                flux_settings_kwargs["language_hints"] = [_to_language_enum(language_code)]
+                if params.get("language_hints"):
+                    flux_settings_kwargs["language_hints"] = params["language_hints"]
+                else:
+                    logger.warning("No Language hint provided. Auto detecting language for Deepgram Flux")
             return DeepgramFluxSTTService(
                 api_key=api_key,
                 sample_rate=SAMPLE_RATE,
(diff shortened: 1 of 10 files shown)


src/eva/assistant/pipeline/services.py
```diff
@@ -320,19 +338,24 @@ def create_tts_service(
...
                 voice=params.get("voice", "aura-2-helena-en"),
-                language=language_code,
+                language=_to_language_enum(language_code),
             ),
         )
 
     elif model_lower == "elevenlabs":
         logger.info(f"Using ElevenLabs TTS: {params['model']}")
+        if (
+            params["model"] not in ("eleven_multilingual_v2", "eleven_flash_v2_5", "eleven_turbo_v2_5")
```
katstankiewicz, line 351: `eleven_multilingual_v2` is multilingual but I think it just doesn't take in the language param

src/eva/assistant/pipeline/services.py
```diff
@@ -320,19 +338,24 @@ def create_tts_service(
...
+        ):
+            raise ValueError(f"ElevenLabs model {params['model']} only supports English language")
         return ElevenLabsTTSService(
             api_key=api_key,
             sample_rate=SAMPLE_RATE,
             settings=ElevenLabsTTSService.Settings(
                 model=params["model"],
                 voice=params.get("voice_id", "hpp4J3VqNfWAUOO0d1Us"),
-                language=language_code,
+                language=_to_language_enum(language_code),
```
```suggestion
                language=_base_language(language_code),
```
elevenlabs doesn't support fr-ca

src/eva/assistant/pipeline/services.py
raghavm243512: It gives a warning on connection but does work. This guard is more of "what will allow it" rather than ElevenLabs specifically accepting the parameter or not. Outside of this, the other models would straight up not be compatible with language (except v3 which is not compatible for other reasons)

katstankiewicz (approved):

PR #144, opened 2026-06-10
Add Deepgram Voice Agent framework support

weiz9:
## Summary

Adds a new `deepgram` assistant-server framework backed by Deepgram's **Voice Agent API** (unified STT→LLM→TTS over a single WebSocket), so the Deepgram agent can be benchmarked like the existing S2S frameworks. Modeled on the Gemini Live server and the `docs/assistant_server_contract.md` contract.

Although it is configured through `s2s_params`, the Voice Agent is a cascade internally (STT→LLM→TTS), so it is **scored as a CASCADE pipeline** — STT WER, transcription accuracy, and speakability metrics all run, and the run_id/folder exposes the three component models (`{stt, llm, tts}`).

## Changes

- **New server** `src/eva/assistant/deepgram_server.py` (`DeepgramAssistantServer`): bridges the Twilio-framed user-simulator WebSocket ↔ `client.agent.v1.connect()`, with the standard audit-log / framework-log / metrics-log / audio-buffer handling.
- **Registration**: `deepgram` added to `worker._get_server_class()` and the `framework` `Literal` in `models/config.py`.
- **Pipeline mapping**: `get_pipeline_type()` maps `deepgram` → `CASCADE` for scoring; `pipeline_parts` exposes `{stt: listen_model, llm: alias-or-model, tts: speak_model}` via the shared `_param_alias()` helper and shared `DEFAULT_LISTEN_MODEL` / `DEFAULT_SPEAK_MODEL` constants.
- **Bring-your-own (BYO) think provider** *(opt-in)*: drive the LLM step with your own credentials/endpoint instead of Deepgram's managed provider — for long prompts that exceed managed limits and for an apples-to-apples comparison with the cascade pipeline. Supports `aws_bedrock` (IAM/STS creds, auto-built from `AWS_*` env vars when omitted), Anthropic/OpenAI/Google via `think_endpoint`, plus `think_params` and `context_length` passthrough. Managed config is unchanged.
- **Custom Google think endpoint**: handles Google's requirement that the model live in the endpoint URL (`.../models/<model>:streamGenerateContent`) and be omitted from settings, by sending the raw settings dict past the SDK's typed-model validation for that case.
- **Fail-loud on think outages**: a provider `Error`, or a greeting-only conversation (caller spoke but the model never replied), is logged at `ERROR` with guidance instead of silently producing a clean-looking 1-turn run.
- **Tests**: `tests/unit/assistant/test_deepgram_server.py` (settings + tool-conversion + BYO-think coverage) and a dispatch case in `test_framework_dispatch.py`.
- **Docs**: Deepgram reference section in `assistant_server_contract.md`; `deepgram` added to the framework enum in `.env.example`.
- **Version**: `simulation_version` 2.0.0 → 2.0.1. This only *adds* a new framework; it does not change any existing framework's outputs, so old simulations do not need re-running.

## Implementation notes (found via live testing)

- **Event parsing**: the server dispatches on the raw WebSocket JSON `type` field rather than the SDK's typed event iterator, whose response-union deserialization mis-routed events (dropping transcripts and tool-call requests) — parsing the JSON directly is deterministic. Binary frames (TTS audio) are handled unchanged.
- **KeepAlive**: a periodic `KeepAlive` task prevents Deepgram's ~10s input-audio timeout from closing the session while the half-duplex agent is speaking.
- **Latency**: `model_response` latency is measured from the server-side receipt time of the simulator's `user_speech_stop` event (the simulator emits it on a monotonic clock).
- **request_id**: the `Welcome` event's `request_id` is logged at `INFO` so each conversation can be traced in Deepgram's server-side logs / support tickets.

### Known limitation — BYO Bedrock model coverage

Deepgram's `aws_bedrock` think integration only supports Claude 3.5-era models (e.g. `anthropic/claude-3-5-haiku-20240307-v1:0`). Haiku 4.5 is rejected with *"Invalid agent.think settings - model not available"*, so BYO Bedrock cannot match the cascade pipeline's Bedrock Haiku 4.5 — use managed or BYO Anthropic-direct (`claude-haiku-4-5`) for Haiku 4.5.

## Config example

```bash
EVA_FRAMEWORK=deepgram
EVA_MODEL__S2S=deepgram
EVA_MODEL__S2S_PARAMS='{"api_key":"<deepgram-key>","model":"gpt-4o-mini"}'
```

Optional `s2s_params` (defaults in parentheses): `think_provider` (`open_ai`), `listen_model` (`nova-3`), `speak_model` (`aura-2-thalia-en`), `alias` (short metrics/run_id label; Deepgram still receives `model`). BYO think (all optional): `think_credentials`, `think_endpoint`, `think_params`, `context_length`, `think_region`. The conversation language comes from the run's `language`, not `s2s_params`.

## Testing

- `ruff` + `ruff format` clean; `pre-commit run --all-files` passes.
- Unit suite passes across assistant / orchestrator / models, including the new Deepgram settings, tool-conversion, framework-dispatch, and BYO-think cases.
- Live end-to-end runs across **airline, itsm, and medical_hr** reached valid conversation ends with tool calls executing scenario-DB mutations and no framework-level errors (disconnects, audio timeouts, or event-processing failures).


src/eva/assistant/deepgram_server.py
```diff
@@ -0,0 +1,532 @@
...
+MULAW_CHUNK_DURATION_S = 0.02  # 20ms per chunk
+
+# Send a KeepAlive at least this often so Deepgram's ~10s input-audio timeout never
+# fires during user silence (e.g. while the agent is speaking).
+KEEPALIVE_INTERVAL_S = 5.0
+
+# Defaults for the Voice Agent listen/think/speak providers (overridable via s2s_params).
+_DEFAULT_LISTEN_MODEL = "nova-3"
+_DEFAULT_THINK_PROVIDER = "open_ai"
+_DEFAULT_THINK_MODEL = "gpt-4o-mini"
```
katstankiewicz:
```suggestion
```
model is a mandatory field so we don't need this default

src/eva/assistant/deepgram_server.py
```diff
@@ -0,0 +1,532 @@
...
+_DEFAULT_SPEAK_MODEL = "aura-2-thalia-en"
+_DEFAULT_LANGUAGE = "en"
```
katstankiewicz:
```suggestion
```
language comes from the config not from model params

src/eva/assistant/deepgram_server.py
```diff
@@ -0,0 +1,532 @@
...
+        s2s_params = self.pipeline_config.s2s_params or {}
+        self._api_key: str = s2s_params.get("api_key", "")
+        # ``think_model`` is the LLM driving the agent; used as the metrics label.
+        # Accept ``model`` as an alias for the contract's "model required" convention.
+        self._think_model: str = s2s_params.get("think_model") or s2s_params.get("model") or _DEFAULT_THINK_MODEL
+        self._model = self._think_model
+        self._think_provider: str = s2s_params.get("think_provider", _DEFAULT_THINK_PROVIDER)
+        self._listen_model: str = s2s_params.get("listen_model", _DEFAULT_LISTEN_MODEL)
+        self._speak_model: str = s2s_params.get("speak_model", _DEFAULT_SPEAK_MODEL)
+        self._language: str = s2s_params.get("language", _DEFAULT_LANGUAGE)
```
katstankiewicz:
```suggestion
```
language is set already in `AbstractAssistantServer`

src/eva/assistant/deepgram_server.py
```diff
@@ -0,0 +1,532 @@
...
+            think["functions"] = self._functions
+
+        settings_dict: dict[str, Any] = {
+            "type": "Settings",
+            "audio": {
+                "input": {"encoding": "linear16", "sample_rate": self._audio_sample_rate},
+                "output": {"encoding": "linear16", "sample_rate": self._audio_sample_rate, "container": "none"},
+            },
+            "agent": {
+                "language": self._language,
```
katstankiewicz:
```suggestion
                "language": self.language,
```

src/eva/assistant/deepgram_server.py
```diff
@@ -0,0 +1,532 @@
...
+        super().__init__(**kwargs)
+
+        # Recording sample rate (Deepgram agent runs at 24 kHz)
+        self._audio_sample_rate = _RECORDING_SAMPLE_RATE
+
+        s2s_params = self.pipeline_config.s2s_params or {}
+        self._api_key: str = s2s_params.get("api_key", "")
+        # ``think_model`` is the LLM driving the agent; used as the metrics label.
+        # Accept ``model`` as an alias for the contract's "model required" convention.
+        self._think_model: str = s2s_params.get("think_model") or s2s_params.get("model") or _DEFAULT_THINK_MODEL
```
katstankiewicz:
```suggestion
        self._think_model: str = s2s_params["model"]
```
we want to use the model set in the params

uv.lock
```diff
@@ -660,7 +660,7 @@ wheels = [
 
 [[package]]
 name = "deepgram-sdk"
-version = "7.3.0"
+version = "7.3.1"
```
JosephMarinier, line 663: Considering this version, does this comment from the PR description still apply?
> SDK workaround: deepgram-sdk 6.1.x mis-deserializes every agent event to the same model via its typed iterator, dropping transcripts and tool-call requests. The server iterates the raw WebSocket and dispatches on the JSON type instead (audio still arrives as binary frames).

src/eva/__init__.py
```diff
@@ -7,7 +7,7 @@
 
 # Bump simulation_version when changes affect benchmark outputs (agent code,
 # user simulator, orchestrator, simulation prompts, agent configs, tool mocks).
-simulation_version = "2.0.1"
+simulation_version = "2.0.5"
```
JosephMarinier: The PR description says this 🤔 
> Version: `simulation_version` 2.0.0 → 2.1.0 (affects benchmark outputs).

Anyway, I'm not sure we need to bump the version. We don't need to rerun any old simulations, right? This addition doesn't really "affect" existing results; it just adds new possibilities. Does that make sense?

```suggestion
simulation_version = "2.0.1"
```

src/eva/models/config.py
```diff
@@ -227,6 +227,17 @@ def pipeline_parts(self) -> dict[str, str]:
...
+                if self.s2s == "deepgram":
+                    # Deepgram Voice Agent is a cascade internally (STT -> LLM -> TTS);
+                    # expose its component models. The `llm` part uses the short
+                    # `think_label` if provided (else the Deepgram model id), so the
+                    # run_id/folder stays readable; defaults mirror deepgram_server.py.
+                    p = self.s2s_params or {}
+                    return {
+                        "stt": p.get("listen_model", "nova-3"),
+                        "llm": p.get("think_label") or p.get("model") or p.get("think_model", ""),
+                        "tts": p.get("speak_model", "aura-2-thalia-en"),
```
JosephMarinier: I'd like to double-check, maybe with @fanny-riols, whether it makes sense to have default models, which will certainly go out of fashion, or if it would be better to require the user to specify the models.

If the default models are OK, we should at least reuse the constants here:
```suggestion
                        "stt": p.get("listen_model", _DEFAULT_LISTEN_MODEL),
                        "llm": p.get("think_label") or p.get("model") or p.get("think_model", ""),
                        "tts": p.get("speak_model", _DEFAULT_SPEAK_MODEL),
```

src/eva/models/config.py
```diff
@@ -227,6 +227,17 @@ def pipeline_parts(self) -> dict[str, str]:
...
                     }
+                if self.s2s == "deepgram":
+                    # Deepgram Voice Agent is a cascade internally (STT -> LLM -> TTS);
+                    # expose its component models. The `llm` part uses the short
+                    # `think_label` if provided (else the Deepgram model id), so the
+                    # run_id/folder stays readable; defaults mirror deepgram_server.py.
+                    p = self.s2s_params or {}
+                    return {
+                        "stt": p.get("listen_model", "nova-3"),
+                        "llm": p.get("think_label") or p.get("model") or p.get("think_model", ""),
```
JosephMarinier: This `"think_model"` is not used anywhere else. Was that a confusion between `"think_label"` and `"think_model"`? Or am I missing something?

Also, is there a reason to use that instead of `"alias"` like in regular S2S models? If we can go with `"alias"`, then we can reuse this function:
```suggestion
                        "llm": _param_alias(p),
```
And adapt `DeepgramAssistantServer.__init__()` accordingly.

src/eva/models/config.py
fanny-riols: Yeah I think it's fine to have defaults for the deepgram framework. But let's use the DEFAULT vars though.

src/eva/models/config.py
```diff
@@ -227,6 +227,17 @@ def pipeline_parts(self) -> dict[str, str]:
...
+                        "tts": p.get("speak_model", "aura-2-thalia-en"),
+                    }
```
JosephMarinier: Thanks to the `_check_companion_services()` model validator, we can assume that `self.s2s_params` is a `dict` (not `None`), so `self.s2s_params or {}` is not necessary.

Combining that with the two suggestions above, we get this. Do all these make sense?
```suggestion
                    return {
                        "stt": self.s2s_params.get("listen_model", _DEFAULT_LISTEN_MODEL),
                        "llm": _param_alias(self.s2s_params),
                        "tts": self.s2s_params.get("speak_model", _DEFAULT_SPEAK_MODEL),
                    }
```

src/eva/assistant/deepgram_server.py
```diff
@@ -0,0 +1,688 @@
...
+        s2s_params = self.pipeline_config.s2s_params or {}
+        self._api_key: str = s2s_params.get("api_key", "")
+        # ``model`` is the exact LLM id sent to Deepgram (required).
+        self._think_model: str = s2s_params["model"]
+        # Metrics/run_id label, decoupled from the (often long) Deepgram model id:
+        # an explicit ``think_label`` if provided, else the model id itself.
+        self._model = s2s_params.get("think_label") or self._think_model
+        self._think_provider: str = s2s_params.get("think_provider", _DEFAULT_THINK_PROVIDER)
+        self._listen_model: str = s2s_params.get("listen_model", _DEFAULT_LISTEN_MODEL)
+        self._speak_model: str = s2s_params.get("speak_model", _DEFAULT_SPEAK_MODEL)
```
JosephMarinier: Instead of introducing `listen_model` and `speak_model` in `s2s_params`, I think it would be possible to use the regular `model` in `stt_params` and `tts_params`. Have you considered that? I'm not sure whether that's a good idea. Let's brainstorm. What do you think?
```suggestion
        self._listen_model: str = (self.pipeline_config.stt_params or {}).get("model", _DEFAULT_LISTEN_MODEL)
        self._speak_model: str = (self.pipeline_config.tts_params or {}).get("model", _DEFAULT_SPEAK_MODEL)
```

src/eva/assistant/deepgram_server.py
```diff
@@ -0,0 +1,688 @@
...
+                "name": tool.function_name,
+                "description": f"{tool.name}: {tool.description}",
+                "parameters": {
+                    "type": "object",
+                    "properties": tool.get_parameter_properties(),
+                    "required": tool.get_required_param_names(),
+                },
+            }
+        )
+    return functions or None
```
JosephMarinier: A list comprehension would be super clear here:
```suggestion
    return [
        {
            "name": tool.function_name,
            "description": f"{tool.name}: {tool.description}",
            "parameters": {
                "type": "object",
                "properties": tool.get_parameter_properties(),
                "required": tool.get_required_param_names(),
            },
        }
        for tool in agent.tools
    ]
```

src/eva/assistant/deepgram_server.py
```diff
@@ -0,0 +1,688 @@
...
+    """Bridges Twilio WebSocket <-> Deepgram Voice Agent API for EVA-Bench evaluation."""
+
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+
+        # Recording sample rate (Deepgram agent runs at 24 kHz)
+        self._audio_sample_rate = _RECORDING_SAMPLE_RATE
+
+        s2s_params = self.pipeline_config.s2s_params or {}
+        self._api_key: str = s2s_params.get("api_key", "")
```
JosephMarinier: Thanks to the `RunConfig._check_companion_services()` model validator, we can assume `"api_key"` is present, so we don't need to guard against the key error. Otherwise, we can let the error be raised.
```suggestion
        self._api_key: str = s2s_params["api_key"]
```

src/eva/assistant/deepgram_server.py
JosephMarinier, line 1: Cool! I don't understand all the implementation details, but that looks very complete. 👍

uv.lock
weiz9: good catch! The description is not updated, I will update it to reflect latest.

src/eva/assistant/deepgram_server.py
weiz9: Yeah, I considered it. Downside is that Deepgram Voice Agent always uses Deepgram for STT/TTS on the same api_key: stt_params/tts_params normally pair with the stt/tts provider fields, so populating them here implies a swap-ability that doesn't exist. Keeping listen_model/speak_model in s2s_params keeps all the Voice Agent config in one validated place. Not strongly attached though, happy to switch if you prefer the consistency. WDYT?

src/eva/models/config.py
weiz9: Done. I moved DEFAULT_LISTEN_MODEL / DEFAULT_SPEAK_MODEL into config.py and imported them into deepgram_server.py, so pipeline_parts and the server now share a single source of truth. (They have to live in config.py rather than the server, config.py can't import the assistant layer without pulling in the heavy deepgram SDK import.) The hardcoded "nova-3"/"aura-2-thalia-en" strings in pipeline_parts are gone. Let me know what you think, thanks!

src/eva/models/config.py
weiz9: Good catch, think_model was a leftover/confusion, it wasn't used anywhere. Done: switched to alias and the branch is now just "llm": _param_alias(p), and I adapted DeepgramAssistantServer.__init__() to read alias too. Also moved the nova-3/aura-2-thalia-en defaults into shared constants. Thanks!

src/eva/models/config.py
weiz9: Makes sense, I dropped the or {} and access self.s2s_params directly (guaranteed non-None by _check_companion_services(), same as the elevenlabs branch). Only tweak: the constants are DEFAULT_LISTEN_MODEL/DEFAULT_SPEAK_MODEL (public, since they're now shared with deepgram_server.py).

src/eva/models/config.py
JosephMarinier: Perfect! Good call on moving the constants to config.py. 👍

src/eva/assistant/deepgram_server.py
JosephMarinier: Good point about the `api_key`; we wouldn't want to have it required in `stt_params` and `tts_params`. But, it wouldn't be, since `_check_companion_services()` won't check `stt_params` and `tts_params` if `pipeline_type` is `S2S`. So I think we're good; we can define `model` (and even `alias`) in `stt_params` and `tts_params` without defining `api_key`.

I'll update my suggestion to make sure we're imagining the same thing.

There are multiple possibilities. I'm not sure which one is the clearest. Maybe it's yours. What do you think?

src/eva/models/config.py
```diff
@@ -227,6 +234,17 @@ def pipeline_parts(self) -> dict[str, str]:
...
+                if self.s2s == "deepgram":
+                    # Deepgram Voice Agent is a cascade internally (STT -> LLM -> TTS);
+                    # expose its component models. The `llm` part uses the short `alias`
+                    # if provided (else the Deepgram model id), so the run_id/folder
+                    # stays readable.
+                    # s2s_params is guaranteed non-None by _check_companion_services().
+                    return {
+                        "stt": self.s2s_params.get("listen_model", DEFAULT_LISTEN_MODEL),
+                        "llm": _param_alias(self.s2s_params),
+                        "tts": self.s2s_params.get("speak_model", DEFAULT_SPEAK_MODEL),
```
JosephMarinier, line 246:
```suggestion
                    # s2s_params is guaranteed non-None by _check_companion_services(), but not stt_params and tts_params.
                    return {
                        "stt": _param_alias(self.stt_params or {}) or DEFAULT_LISTEN_MODEL,
                        "llm": _param_alias(self.s2s_params),
                        "tts": _param_alias(self.tts_params or {}) or DEFAULT_SPEAK_MODEL,
```

src/eva/models/config.py
```diff
@@ -51,6 +51,13 @@ def _get_all_metrics() -> list[str]:
...
+# Defaults for the Deepgram Voice Agent listen/speak providers (overridable via s2s_params).
+# Defined here (rather than in the assistant layer) so pipeline_parts can reference them
+# without importing deepgram_server.py — which would pull the heavy deepgram SDK import.
+DEFAULT_LISTEN_MODEL = "nova-3"
+DEFAULT_SPEAK_MODEL = "aura-2-thalia-en"
+
+
 def _param_alias(params: dict[str, Any]) -> str:
     """Return the display alias from a params dict."""
     return params.get("alias") or params["model"]
```
JosephMarinier, line 63:
```suggestion
    return params.get("alias") or params.get("model")
```

src/eva/assistant/deepgram_server.py
```diff
@@ -0,0 +1,687 @@
...
+        s2s_params = self.pipeline_config.s2s_params or {}
+        self._api_key: str = s2s_params["api_key"]
+        # ``model`` is the exact LLM id sent to Deepgram (required).
+        self._think_model: str = s2s_params["model"]
+        # Metrics/run_id label, decoupled from the (often long) Deepgram model id:
+        # an explicit ``alias`` if provided, else the model id itself.
+        self._model = s2s_params.get("alias") or self._think_model
+        self._think_provider: str = s2s_params.get("think_provider", _DEFAULT_THINK_PROVIDER)
+        self._listen_model: str = s2s_params.get("listen_model", DEFAULT_LISTEN_MODEL)
+        self._speak_model: str = s2s_params.get("speak_model", DEFAULT_SPEAK_MODEL)
```
JosephMarinier, line 113:
```suggestion
        self._listen_model: str = (self.pipeline_config.stt_params or {}).get("model", DEFAULT_LISTEN_MODEL)
        self._speak_model: str = (self.pipeline_config.tts_params or {}).get("model", DEFAULT_SPEAK_MODEL)
```

JosephMarinier (commented):
Here's one possibility: reuse the existing `stt/tts_params.get("model")` (and even `stt/tts_params.get("alias")` for the config's `pipeline_parts`).

src/eva/models/config.py
JosephMarinier, line 246:
```suggestion
                        "stt": self.stt or DEFAULT_LISTEN_MODEL,
                        "llm": _param_alias(self.s2s_params),
                        "tts": self.tts or DEFAULT_SPEAK_MODEL,
```

src/eva/assistant/deepgram_server.py
JosephMarinier, line 113:
```suggestion
        self._listen_model: str = self.pipeline_config.stt or DEFAULT_LISTEN_MODEL
        self._speak_model: str = self.pipeline_config.tts or DEFAULT_SPEAK_MODEL
```

JosephMarinier (commented):
And here's another even simpler possibility: reuse the existing `stt/tts`.

PR #147, opened 2026-06-11
loc translations, integrity test, token approx

raghavm243512:
Force user goals to see translated location names so they don't say things in English
Database integrity test: constructs expected DB from initial DB and expected trace for every language. Everything from names, phone numbers, and location names are all translated per language during the replay

Threw in a small fallback for LLM and alm_vllm reasoning token count as well, and added compatibility for latest vllm versions


data/itsm_dataset.json
```diff
@@ -3437,6 +3437,42 @@
             "event_type": "user_utterance",
             "utterance": "Yes, please submit it."
           },
+          {
```
gabegma, line 3440: It might not matter a lot, but these tool calls are not in the right place in the expected trace right?

src/eva/assistant/pipeline/alm_vllm.py
```diff
@@ -122,6 +124,20 @@ async def complete(
                     if details and hasattr(details, "reasoning_tokens"):
                         reasoning_tokens = getattr(details, "reasoning_tokens", 0)
 
+                if reasoning_content and reasoning_tokens == 0:
+                    if not getattr(self, "_reasoning_token_fallback_warned", False):
+                        logger.warning(
+                            "No reasoning token count found in API response for model '%s'; "
+                            "falling back to tiktoken approximation. This warning will not repeat.",
+                            self.model,
+                        )
```
gabegma:
```suggestion
                        logger.warning(
                            f"No reasoning token count found in API response for model {self.model}; "
                            "falling back to tiktoken approximation. This warning will not repeat.",
                        )
```

src/eva/assistant/services/llm.py
```diff
@@ -119,6 +120,19 @@ async def complete(
                 if reasoning_content:
                     logger.info(f"💭 Reasoning content from {model} ({len(reasoning_content)} chars)")
                     logger.debug(f"Reasoning content preview: {reasoning_content[:200]}...")
+                    if reasoning_tokens == 0:
+                        if not getattr(self, "_reasoning_token_fallback_warned", False):
```
gabegma: Can you define it in the __init__ to `False` instead of doing this?

data/itsm_dataset.json
raghavm243512: yea those were hand filled, the trace is really only used to extract and apply write tools

src/eva/assistant/services/llm.py
```diff
@@ -119,6 +120,19 @@ async def complete(
                 if reasoning_content:
                     logger.info(f"💭 Reasoning content from {model} ({len(reasoning_content)} chars)")
                     logger.debug(f"Reasoning content preview: {reasoning_content[:200]}...")
+                    if reasoning_tokens == 0:
```
gabegma, line 124: A utility function would be nice since the code is the same in both classes.

src/eva/utils/culture.py
```diff
@@ -54,6 +58,49 @@
...
+
+    English (and any language without a translation) renders the canonical English
+    ``name``. Other languages render the alias's primary translation
+    (``translations[language][0]``). Unknown alias names are left as the literal name.
+    """
+
+    def render(name: str) -> str:
+        entry = index.get(name)
+        if entry is None:
+            logger.warning("LOC placeholder references unknown alias %r; using literal name.", name)
```
gabegma:
```suggestion
            logger.warning(f"LOC placeholder references unknown alias {claude}; using literal name.")
```

src/eva/utils/culture.py
```diff
@@ -54,6 +58,49 @@
...
+            return name
+        if _is_english(language):
+            return name
+        translations = (entry.get("translations") or {}).get(language) or []
+        if not translations:
+            logger.warning("Alias %r has no %r translation; rendering English name in user goal.", name, language)
```
gabegma:
```suggestion
            logger.warning(f"Alias {name} has no {language} translation; rendering English name in user goal.")
```

src/eva/utils/culture.py
```diff
@@ -169,9 +225,12 @@ def resolve_user_goal(
             f"Run scripts/add_culture_data.py --language {language} to populate it."
         )
     utt = starting_utterances[language]
-    resolved["starting_utterance"] = _replace_in(
-        utt, first, last, first_rom, last_rom, phone, comp_first, comp_first_rom
-    )
+    resolved_utt = _replace_in(utt, first, last, first_rom, last_rom, phone, comp_first, comp_first_rom)
```
gabegma: This is not new - but it would be better to add * in the args of `_replace_in`, and to add keywords arguments here, since it would be easy to mess up the order here.

src/eva/utils/culture.py
```diff
@@ -169,9 +225,12 @@ def resolve_user_goal(
...
+    if aliases_dir is not None:
+        index = _load_aliases_index(str(aliases_dir))
+        if index:
+            resolved_utt = _resolve_locations(resolved_utt, index, language)
```
gabegma: Could we do this in `_replace_in` to not duplicate the code in both places?

tests/unit/data/test_db_integrity.py
```diff
@@ -0,0 +1,556 @@
...
+
+
+# ---------------------------------------------------------------------------
+# End-to-end DB integrity replay
+# ---------------------------------------------------------------------------
+
+REPO_ROOT = Path(__file__).resolve().parents[3]
+
+REPLAY_DOMAINS = ["itsm", "medical_hr", "airline"]
+RESOLVE_DOMAINS = ["itsm", "medical_hr", "airline"]
```
gabegma, line 273: Could we move this out in a new file - and add this under `tests/unit/data.py` or something like this?

tests/unit/utils/test_multilingual_integrity.py
```diff
@@ -0,0 +1,556 @@
...
+
+def _resolve_params(params, culture_overrides, language, romanized_culture_overrides, aliases_index=None):
+    """Resolve name and location placeholders in tool call params.
+
+    - ``<FIRST_NAME>`` / ``<LAST_NAME>`` / ... → per-language name values.
+    - ``<LOC:Name>`` → language-appropriate alias (e.g. "campus est" in French),
+      so the tool receives the translated form and exercises the name_aliases
+      lookup chain end-to-end.
+    """
+    from eva.utils.culture import _companion_for, _names_for, _phone_for
```
gabegma: Put at the top

tests/unit/data/test_db_integrity.py
```diff
@@ -0,0 +1,556 @@
...
+
+def _load_dataset(domain: str) -> list[dict]:
+    return json.loads(_domain_paths(domain)["dataset"].read_text())
+
+
+def _load_scenario(domain: str, record_id) -> dict:
+    return json.loads((_domain_paths(domain)["scenarios"] / f"{record_id}.json").read_text())
+
+
+def _resolve_params(params, culture_overrides, language, romanized_culture_overrides, aliases_index=None):
```
gabegma, line 294: Is it not possible to use the real code here, instead of re-implementing?  So we can test the real thing?

tests/unit/data/test_db_integrity.py
```diff
@@ -0,0 +1,556 @@
...
+    paths = _domain_paths(domain)
+    return ToolExecutor(
+        tool_config_path=str(paths["agent_config"]),
+        scenario_db_path=str(paths["scenarios"] / f"{sample_record_id}.json"),
+        tool_module_path=paths["tool_module"],
+        current_date_time="2026-01-01 00:00 EST",
+    )
+
+
+def _apply_subs(obj, subs: dict[str, str]):
```
gabegma, line 312: Same for this

gabegma (approved):
Excellent work Raghav - thanks for taking the time to make this solution more robust and fail proof with a test!

data/itsm_dataset.json
gabegma: I'm just nervous people might use them for other purposes, so ideally, the tool calls would still be correctly placed.

tests/unit/data/test_db_integrity.py
raghavm243512: this one is test (used on trace) only - no equivalent used in EVA sims. But other one is fixed

PR #152, opened 2026-06-16
Use Flex processing for the OpenAI judge model

JosephMarinier:
Use Flex processing for the OpenAI judge model, which will halve its cost in exchange for slower response times and occasional resource unavailability. This doesn't affect benchmarking an OpenAI model. More details on flex tier [here](https://developers.openai.com/api/docs/guides/flex-processing).

### DIFF
--- a/src/eva/metrics/accuracy/faithfulness.py
+++ b/src/eva/metrics/accuracy/faithfulness.py
@@ -60,6 +60,7 @@ class FaithfulnessJudgeMetric(ConversationTextJudgeMetric):
     )
     category = "accuracy"
     default_model = "us.anthropic.claude-opus-4-6-v1"
+    default_params = {"max_tokens": 100000}  # Drop the OpenAI-only flex tier inherited from TextJudgeMetric.
     rating_scale = (1, 3)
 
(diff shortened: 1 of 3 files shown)


src/eva/metrics/accuracy/faithfulness.py
```diff
@@ -60,6 +60,7 @@ class FaithfulnessJudgeMetric(ConversationTextJudgeMetric):
     )
     category = "accuracy"
     default_model = "us.anthropic.claude-opus-4-6-v1"
+    default_params = {"max_tokens": 100000}  # Drop the OpenAI-only flex tier inherited from TextJudgeMetric.
```
fanny-riols, line 63: are we sure 100k is enough? Did you check the prompt size after 10 min convo?

JosephMarinier: I did not check that myself, but it has been 100k since the beginning of EVA (inherited from `src/eva/metrics/base.py`). Is that OK?

fanny-riols: ah right, I guess so then.

fanny-riols (approved):

PR #156, opened 2026-06-19
Add domain in default run_id

JosephMarinier:
Add the domain name in the default `run_id`.

### DIFF
--- a/src/eva/models/config.py
+++ b/src/eva/models/config.py
@@ -655,7 +655,8 @@ class RunConfig(BaseSettings):
         if "run_id" not in self.model_fields_set:
             suffix = "_".join(v for v in self.model.pipeline_parts.values() if v)
             lang = self.language.value
-            self.run_id = f"{datetime.now(UTC):%Y-%m-%d_%H-%M-%S.%f}_{self.domain}_{lang}_{suffix}"
+            domain = self.domain.replace("_", "-")
+            self.run_id = f"{datetime.now(UTC):%Y-%m-%d_%H-%M-%S.%f}_{domain}_{lang}_{suffix}"
 
         return self
 
--- a/src/eva/models/config.py
+++ b/src/eva/models/config.py
@@ -655,7 +655,7 @@ class RunConfig(BaseSettings):
         if "run_id" not in self.model_fields_set:
             suffix = "_".join(v for v in self.model.pipeline_parts.values() if v)
             lang = self.language.value
-            self.run_id = f"{datetime.now(UTC):%Y-%m-%d_%H-%M-%S.%f}_{lang}_{suffix}"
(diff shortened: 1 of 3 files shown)


katstankiewicz (approved):
ty!!

src/eva/models/config.py
```diff
@@ -655,7 +655,7 @@ def _check_companion_services(self) -> "RunConfig":
         if "run_id" not in self.model_fields_set:
             suffix = "_".join(v for v in self.model.pipeline_parts.values() if v)
             lang = self.language.value
-            self.run_id = f"{datetime.now(UTC):%Y-%m-%d_%H-%M-%S.%f}_{lang}_{suffix}"
+            self.run_id = f"{datetime.now(UTC):%Y-%m-%d_%H-%M-%S.%f}_{self.domain}_{lang}_{suffix}"
```
JosephMarinier: My only concern is that the domain name `medical_hr` contains an underscore, which is the separator we use between the bits of `run_id` here. A `run_id` like `<date>_medical_hr_en_<models>` is slightly confusing. Furthermore, `hr` is the language code for Croatian.

A couple of solutions:
1. Write `medical-hr` instead of `medical_hr`. There are often dashes within language codes and within model names. But then, it doesn't match the domain name exactly.
2. Change our separator for something that risks fewer collisions, like a tilde `~`.
3. Live with it.
4. Not add the domain name (close this PR).

What do you think?

katstankiewicz: hmm, I really like the addition of domain. what about using `medicalhr`? it won't match domain exactly but reduces the confusion with language codes and needed a new separator. I'm ok with `medical-hr` as well though

JosephMarinier: That's a good idea too!

What I like about replacing the underscore (rather than removing it) is that it's a reversible transformation; we're not losing any info.
```suggestion
            domain = self.domain.replace("_", "-")
            self.run_id = f"{datetime.now(UTC):%Y-%m-%d_%H-%M-%S.%f}_{domain}_{lang}_{suffix}"
```

src/eva/orchestrator/runner.py
```diff
@@ -276,8 +292,15 @@ async def _run_and_pipeline(
...
 
-            pipeline_results = await asyncio.gather(
+            pipeline_results = await tqdm.gather(
                 *(_run_and_pipeline(output_id_to_record[oid], oid) for oid in pending_output_ids),
+                total=len(pending_output_ids),
+                desc=f"Attempt {attempt_number}/{max_attempts}",
+                file=TqdmNewlineStream(sys.stderr),
+                mininterval=0,
+                unit="task",
+                smoothing=0,  # Smoothing would produce unstable estimates because tasks run in parallel.
```
JosephMarinier, line 302: Image

PR #160, opened 2026-06-22
Pr/config app updates

raghavm243512:
save inactive entries, fix text only run bug


weiz9 (approved):
Looks good, thanks raghav

PR #170, opened 2026-06-30
gemini user sim and audio quality

raghavm243512:
adds gemini user simulator and clears some electrical static when openai is used as the backend assistant


raghavm243512:
Drop this in favor of new architecture

PR #174, opened 2026-07-02
Add SmallestAI Pulse STT and Lightning TTS to pipecat cascade pipeline

weiz9:
## Summary
- Wires pipecat's `SmallestSTTService` (Pulse) and `SmallestTTSService` (Lightning v3.1 / v3.1 Pro) into `create_stt_service`/`create_tts_service`, so the CASCADE pipeline (STT -> LLM -> TTS) can use Smallest as a provider alongside the existing SmallestAI Hydra S2S integration.
- Requires `pipecat-ai>=1.4.0` (already the pinned minimum in `pyproject.toml`).
- Bumped `simulation_version` (2.0.1 -> 2.0.2) since this touches `src/eva/assistant/`.
- Added `smallest` to the STT/TTS provider enums in `.env.example`.

## Test plan
- [x] `pytest tests/unit/assistant/test_services.py -v` — new smallest STT/TTS unit tests pass, including the default-voice-preservation edge case (voice_id omitted -> pipecat's per-model default voice applies).
- [x] `pytest tests/unit -q` — full unit suite passes (1846 passed).
- [x] `ruff check` / `ruff format --check` clean; `mypy` shows no new errors beyond the pre-existing `api_key: Any | None` pattern shared by every other provider in this file.
- [x] End-to-end `EVA_DEBUG=true eva --domain airline` with `stt=smallest` (pulse), `tts=smallest` (lightning_v3.1_pro), `llm=gpt-5.4`: 3/3 conversations succeeded. Verified valid non-silent 24kHz audio, clean multi-turn transcripts, zero error frames from `SmallestSTTService`/`SmallestTTSService` in pipecat logs, and reasonable STT WER (~9%).

### DIFF
--- a/src/eva/__init__.py
+++ b/src/eva/__init__.py
@@ -7,7 +7,7 @@ __version__ = "2.0.0"
 
 # Bump simulation_version when changes affect benchmark outputs (agent code,
 # user simulator, orchestrator, simulation prompts, agent configs, tool mocks).
-simulation_version = "2.0.2"
+simulation_version = "2.0.1"
 
 # Bump metrics_version when changes affect metric computation (metrics code,
 # judge prompts, pricing tables, postprocessor).
--- a/src/eva/__init__.py
+++ b/src/eva/__init__.py
@@ -7,7 +7,7 @@ __version__ = "2.0.0"
 
 # Bump simulation_version when changes affect benchmark outputs (agent code,
 # user simulator, orchestrator, simulation prompts, agent configs, tool mocks).
-simulation_version = "2.0.1"
+simulation_version = "2.0.2"
 
 # Bump metrics_version when changes affect metric computation (metrics code,
 # judge prompts, pricing tables, postprocessor).
--- a/src/eva/assistant/pipeline/services.py
+++ b/src/eva/assistant/pipeline/services.py
@@ -282,7 +282,7 @@ def create_stt_service(
         return stt_service
 
     elif model_lower == "smallest":
-        logger.info(f"Using Smallest Pulse STT: {params['model']}")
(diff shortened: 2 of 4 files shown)


src/eva/__init__.py
```diff
@@ -7,7 +7,7 @@
 
 # Bump simulation_version when changes affect benchmark outputs (agent code,
 # user simulator, orchestrator, simulation prompts, agent configs, tool mocks).
-simulation_version = "2.0.1"
+simulation_version = "2.0.2"
```
fanny-riols: I don't think we need to bump the version here. You are only adding support to smallestai, so it doesn't affect the simulator.
```suggestion
simulation_version = "2.0.1"
```

src/eva/assistant/pipeline/services.py
```diff
@@ -279,6 +281,21 @@ def create_stt_service(
             )
         return stt_service
 
+    elif model_lower == "smallest":
+        logger.info(f"Using Smallest Pulse STT: {params['model']}")
```
fanny-riols: The model name might change with time.
```suggestion
        logger.info(f"Using Smallest STT: {params['model']}")
```

src/eva/assistant/pipeline/services.py
```diff
@@ -500,6 +517,25 @@ def create_tts_service(
 
         return openai_tts
 
+    elif model_lower == "smallest":
+        logger.info(f"Using Smallest Lightning TTS: {params['model']}")
```
fanny-riols:
```suggestion
        logger.info(f"Using Smallest TTS: {params['model']}")
```

fanny-riols (approved):

PR #180, opened 2026-07-10
Extend key entities metric to audio-native models

weiz9:
## What

Extends the existing `transcription_accuracy_key_entities` metric to support **audio-native** pipelines (S2S / AUDIO_LLM). It stays a single text-to-text judge — no new metric class.

Previously the metric was CASCADE-only: it compared what the user was instructed to say (`intended_user_turns`) against what the agent's STT transcribed (`transcribed_user_turns`). Audio-native models expose no reliable STT transcript, so entity understanding was **unmeasured** for them.

## How

The metric now branches on pipeline type:

- **CASCADE** (unchanged): compares `intended_user_turns` vs `transcribed_user_turns` (STT output) — transcription accuracy.
- **Audio-native** (new): compares the user's intended entities against the agent's **tool-call arguments** (from `conversation_trace`), which are the evidence of what the model actually understood. If the user says "my id is EM123" and the agent calls a tool with "EMP124", the entity is flagged as misheard.

Both paths reuse the same per-entity ratio scoring (`correct / total`) and per-entity-type sub-metrics. Audio-native uses a new `audio_native_prompt` judge template; the cascade `user_prompt` is untouched.

## Changes

- `transcription_accuracy_key_entities.py`: removed the CASCADE-only restriction (now all pipeline types), added `_build_prompt_and_turns()` to select source + prompt via `context.is_audio_native`, plus `_get_audio_native_turns()` / `_format_intended_user_turns()`. Bumped metric `version` v0.3 → v0.4.
- `configs/prompts/judge.yaml`: added `audio_native_prompt` under the same metric key.
- Tests: added `TestPipelineSupport` and `TestAudioNativeCompute` (misheard → 0, correct → 1, verifies the audio-native prompt is used with tool-call evidence).
- Bumped `metrics_version` 2.2.0 → 2.3.0 and regenerated `metric_signatures.json`.

## Validation

**Unit tests:** 17 passing (existing cascade cases + new audio-native cases); full metrics suite green (423), including the drift/signature test. ruff clean.

**Audio-native behavioral check** — real AUDIO_LLM run (`nowllm-speech`), 5 conversations × 3 trials:
- Metric now activates on audio-native runs (was silently filtered out before), mean score **0.92**, no errors.
- Correctly routed to the `audio_native_prompt` and compared user entities against tool-call arguments.
- All 11 incorrect verdicts were **genuine mishears**, e.g. conf. code `ZK3FFW`→`VK3FFW`, `6VORJU`→`6BORTU`, name `Thompson`→`Lee`, destination `Washington DC`→`DFW (Dallas)`. No false positives on correctly-captured entities.

**Cascade regression** — 3 cascade runs (xai-stt, flux, pulse), 9 trials:
- Every generated judge prompt is **byte-identical** to the prior v0.3 output → cascade input path provably unchanged.
- All trials routed to the cascade prompt (never audio-native); identical entity/eval counts; no errors.
- 7/9 scores identical; 2 differ by ~0.04–0.06 — attributable to LLM judge nondeterminism (identical inputs, judge runs without fixed temperature), not code.

## Notes

- **Open item (minor, non-blocking):** the audio-native judge currently treats soft constraints ("under $80", "before 2 PM") as entities and marks them correct when the agent's action satisfies them. Defensible, but broadens scope beyond literal identifier capture — a prompt-tuning tweak if we want it limited to hard identifiers (codes, names, IDs).
- **Signature coverage gap:** the drift signature only hashes `judge.{name}.user_prompt`, so future text edits to the new `audio_native_prompt` won't auto-trip the drift test — they'd need a manual version bump. Happy to extend `signatures.py` to hash all prompt keys if preferred.
- Pre-existing mypy errors in this file (MetricScore pydantic defaults under `strict`) are unchanged; the new code adds none.


configs/prompts/judge.yaml
```diff
@@ -1071,3 +1071,169 @@ judge:
...
+      - The agent using an entity that came from a tool RESPONSE rather than from the user (only user-provided entities are evaluated here)
+      - Correct values that simply never got reused (mark those skipped, not incorrect)
+
+      Only mark an entity incorrect when the agent demonstrably used a **different value** than the user provided (e.g. user "EM123" → tool call "EMP124", user "March 25th" → tool call "2026-03-26").
+
+      ## Understanding Tags in the User Text
+      The user text may contain non-spoken tags and markers. These are metadata — they were never said aloud and must not be treated as entities or evaluated.
+
+      ### Audio-Direction Tags
+      Tags like [slow], [firm], [annoyed] describe how the words were meant to be spoken. Ignore them entirely.
```
gabegma: Are those the same thing? Should they be combined?

configs/prompts/judge.yaml
```diff
@@ -1071,3 +1071,169 @@ judge:
...
+      These markers indicate that parts of the user text may never have been spoken aloud, because the user was interrupted or talked over. An entity that was never spoken cannot have been understood, so you must NOT penalize for entities in regions that were likely not said — mark them as skipped.
+
+      **Key principle:** Only evaluate entities that were reasonably expected to have been spoken aloud AND that the agent had an occasion to use. If a tag indicates the user was interrupted before or during an entity, still include the entity but set `skipped: true` and explain why.
+
+      ## User Turns to Evaluate
+      {user_turns}
+
+      ## Conversation (with agent turns and tool calls/responses)
+      Use this to find where each user-provided entity was used by the agent. Tool call arguments are the primary evidence of what the agent understood.
+      {conversation}
```
gabegma, line 1121: I wonder if we should redact the assistant turns here - I'm nervous if there is a transcription error in them, the judge could use those to penalize key entities. For example, if the user says "INC462" and the agent replies, "Did you say INC463"? The latter could just be a transcription error, so it should not be penalized. I think the judge should only see user turns and tool calls/responses, just to avoid this issue. Also, could we show the user turns in the conversation directly, instead of here and above? The `conversation` could be user turns, assistant turns (marked as redacted, as I think we do in speech fidelity), and tool calls/responses.

configs/prompts/judge.yaml
```diff
@@ -1071,3 +1071,169 @@ judge:
...
+      - Non-Latin scripts: A name or identifier rendered in its native writing system (e.g. Kanji, Hangul, Arabic) is equivalent to its romanized/phonetic form if they represent the same spoken entity. For example, "山田" and "Yamada" should be treated as matching.
+      - Phonetic letter spellings: In languages that do not use the Latin alphabet, Roman/Latin letters are spoken using language-specific phonetic equivalents. These are correct, not errors (e.g., in Korean "에이" for "A" or "비" for "B").
+      - Number surface variants: Many languages have systematic spoken-form differences that do not change the underlying value and must not be penalized:
+        - Reversed digit order in spoken form (e.g., German "vierundzwanzig" = four-and-twenty = 24)
+        - Grammatical gender/agreement variants (e.g., Spanish "veintiún" vs "veintiuna" for 21)
+        - Large-unit grouping differences (e.g., Japanese 万-system: "ichi-man go-sen" / 一万五千 = 15,000)
+      - Honorific/title variants: "Ms." vs "Miss", "Mr." vs "Mister", and language equivalents (e.g., French "Madame" vs "Mme", Japanese "様" vs "-san") are acceptable variants of the same title.
```
gabegma: Would those be key entities? The Honorific/title variants part.

gabegma (commented):
Nice! This looks good! Only a few comments. Have you tested it? Edit: nevermind, just read the description, cool!!

gabegma:
Oh, and yes, could we fix signatures, so that changes in this new prompt are also caught? Thank you!

configs/prompts/judge.yaml
weiz9: They're two distinct kinds with different handling, but I will combine them into a single "Non-Spoken Tags in the User Text" section.

configs/prompts/judge.yaml
weiz9: good catch, I agree on both. I've made changes so now the judge now gets a single redacted Conversation (assistant turns shown as [Assistant speaks], user turns inline) so mis-transcribed agent read-backs like your INC462/INC463 case can't penalize an entity. Let me know what you think!

configs/prompts/judge.yaml
weiz9: Agreed, I removed it, a standalone title isn't a concrete entity, so I moved honorifics to the explicit "Not an entity" list.

weiz9:
> Oh, and yes, could we fix signatures, so that changes in this new prompt are also caught? Thank you!

Fixed! signatures.py now hashes all prompt keys under judge.{name}, so audio_native_prompt edits are caught by the drift test.

configs/prompts/judge.yaml
```diff
@@ -1075,13 +1075,15 @@ judge:
...
+      This is a speech-to-speech / audio-native agent: there is no reliable speech-to-text transcript of the user to inspect. Instead, the evidence of what the agent actually understood is the **arguments it passed to its tool calls**. When a user says "my id is EM123" and the agent then calls a tool with "EMP124", the agent misheard the entity.
+
+      You are given a Conversation containing user turns, tool calls, and tool responses. **The assistant's own spoken turns are redacted** (shown as `[Assistant speaks]`) and deliberately withheld: the agent's speech is itself imperfectly transcribed, so it must never be used as evidence. For example, if the user says "INC462" and the agent asks "Did you say INC463?", that read-back may just be a transcription artifact — do not treat it as the agent mishearing the entity. Judge only what the user said (user turns) against what the agent did (tool-call arguments).
```
gabegma: I think we can drop the last part since it's redacted, starting at `For example,`

configs/prompts/judge.yaml
```diff
@@ -1075,13 +1075,15 @@ judge:
...
 
       Your task:
-      1. For EACH user turn, identify all key entities in what the USER SAID (the "User said" text).
-      2. For each entity, find where the agent used it — a tool call argument (see the Conversation below) or a value the agent read back — and check whether the agent captured it CORRECTLY.
+      1. For EACH user turn in the Conversation, identify all key entities in what the USER SAID.
+      2. For each entity, find whether the agent used it in a tool-call argument and check whether it was captured CORRECTLY.
```
gabegma: What happens if it's used in multiple tool calls? Should we mention this?

gabegma: I think one issue is that the key entity might come up from the user at a specific turn, but the tool calls with that entity might only come in later turns. So it might be unintuitive for the judge to have to check all future tool calls when determining whether a specific entity is correct.

Even worse, imagine turn 1 has "INC345" - but the user actually made a mistake.
Turn 2 has "INC436" - the user says it got confused, fixes its mistake. This is the correct entity.
Turn 4 has a tool call with INC435 (wrong)
Turn 8 has a tool call with INC436 (correct)

It would be weird for the judge to score this one with the existing framing - is turn 1 correct or no? Is turn 2 correct or no? It might be more intuitive to ask for a rating, not when an entity is said, but rather, when a tool call containing an entity is made. So, turn 1 and 2 here would be skipped, but the judge would provide a rating for turn 4 and 8 instead. That would also allow for a key entity to be correct in one tool call and wrong in the other. 

Does that make sense? Not sure if I explained it well.

configs/prompts/judge.yaml
gabegma: Super - looks good!

tests/unit/metrics/test_transcription_accuracy_key_entities.py
```diff
@@ -431,6 +431,43 @@ async def test_uses_audio_native_prompt_and_intended_turns(self, metric):
...
+        )
+        generate_text = AsyncMock(return_value=(response, None))
+        metric.llm_client.generate_text = generate_text
+
+        await metric.compute(context)
+
+        sent_prompt = generate_text.call_args[0][0][0]["content"]
+        assert "[Assistant speaks]" in sent_prompt  # assistant turn present but redacted
+        assert "QX7711" not in sent_prompt  # the mis-transcribed read-back is withheld
+        assert "QX7710" in sent_prompt  # user turn + tool call are visible
```
gabegma: Detail - but this is not checking that both user turn and tool calls are visible

configs/prompts/judge.yaml
weiz9: Done, dropped that sentence since the assistant turns are redacted and the judge never sees the read-back anyway.

configs/prompts/judge.yaml
weiz9: Yes! That makes a lot of sense too me, good catch as well! Reframed exactly as you suggested: ratings are now keyed per tool call (Tool Call [N]), user-only turns aren't rated, and the same value can be wrong in one call and right in another. Validated on a live audio-native run where it correctly rated each of the reservation/rebook/seat calls separately.

tests/unit/metrics/test_transcription_accuracy_key_entities.py
weiz9: Fixed, the test now asserts both the user-turn line and the tool-call line are present in the prompt separately.

configs/prompts/judge.yaml
```diff
@@ -1077,14 +1077,16 @@ judge:
...
+      2. Within each tool call, look at **every argument**. For any argument whose value traces back to something the USER said — a user-provided key entity (name, confirmation code, date, amount, seat preference, etc.) — work out what the user actually said for it.
+      3. Rate whether the value in that tool call matches what the user said: correct or incorrect.
+      4. A user often states an entity in one turn and the agent uses it in a tool call several turns later — always rate it at the tool call, never at the user turn. If the same user value is used in more than one tool call, rate it in EACH one (so it can be correct in one tool call and wrong in another).
+      5. If a tool call uses no user-provided entities (every argument came from a tool response or was generated by the agent), omit it from your output entirely.
+      6. Only evaluate arguments that trace back to a value the USER provided. Ignore arguments the agent got from a tool RESPONSE or generated itself (e.g. a journey/record id returned by a prior search, or a passenger id) — those are not user entities.
```
gabegma: This seems the same thing?

configs/prompts/judge.yaml
```diff
@@ -1077,14 +1077,16 @@ judge:
...
+      7. If a user-provided entity falls in a region that was likely never spoken aloud (interruption tags), mark it skipped and explain why.
```
gabegma: I don't think this is needed anymore - it won't show up in a tool call if it was never said.

configs/prompts/judge.yaml
```diff
@@ -1140,85 +1142,72 @@ judge:
...
-          "turn_id": 1,
+          "tool_call_id": 1,
+          "tool_name": "lookup_employee",
           "entities": [
           {{
               "type": "employee_id",
-              "value": "E M one two three",
-              "captured_value": "EMP124",
-              "analysis": "User said EM123 but the agent called lookup_employee with EMP124 — misheard.",
+              "value": "E M four five six (user corrected it in turn 2)",
```
gabegma: I don't think we want the explanation in the value, would be better in analysis only. So only `E M four five six`

configs/prompts/judge.yaml
```diff
@@ -1140,85 +1142,72 @@ judge:
...
-      Turn 1 - Tool Call (lookup_employee): {{'employee_id': 'EMP124'}}
-      Turn 2 - User: Book it for December 15th.
+      Turn 2 - User: Actually, sorry — the id is E M four five six.
       Turn 2 - [Assistant speaks]
-      Turn 2 - Tool Call (create_booking): {{'date': '2026-12-15'}}
-      Turn 3 - User: [slow] The code is X X F six O H, with the letter O, [assistant interrupts] not zero.
-      Turn 3 - [Assistant speaks]
+      Turn 4 - [Assistant speaks]
+      Turn 4 - Tool Call [1] (lookup_employee): {{'employee_id': 'EM455'}}
+      Turn 4 - Tool Response (lookup_employee): {{'employee_id': 'EM455', 'record_id': 'R-9012'}}
```
gabegma: Perhaps for more realism, we could put EM123 here - and show to the judge that this would also be wrong, given the user had already corrected it.

configs/prompts/judge.yaml
```diff
@@ -1140,85 +1142,72 @@ judge:
...
               "type": "<name|date|time|confirmation_code|flight_number|amount|address|phone|email|employee_id|etc...>",
-              "value": "<entity value as the user said it>",
-              "captured_value": "<the value the agent used in a tool call, or 'not used'>",
-              "analysis": "<brief reason; if skipped, explain why (never used by the agent, or in an interrupted region)>",
+              "value": "<what the user actually said for this entity, accounting for any later correction>",
+              "captured_value": "<the value this tool call passed>",
+              "analysis": "<brief reason for the correct/incorrect judgement>",
               "correct": <true|false>,
-              "skipped": <true|false>
+              "skipped": <true|false: true only for entities in an interruption region>
```
gabegma: I don't think skipped still applies now?

tests/fixtures/metric_signatures.json
```diff
@@ -85,9 +85,9 @@
   },
   "TranscriptionAccuracyKeyEntitiesMetric": {
     "name": "transcription_accuracy_key_entities",
-    "prompt_hash": "3ed3a2fe7dcb",
-    "source_hash": "2926af63f4da",
-    "version": "v0.5"
+    "prompt_hash": "778c999bc0e5",
+    "source_hash": "adf7083be05b",
+    "version": "v0.6"
```
gabegma: No need to bump more than once in one PR.

src/eva/metrics/diagnostic/transcription_accuracy_key_entities.py
```diff
@@ -114,7 +114,14 @@ async def compute(self, context: MetricContext) -> MetricScore:
                     error=error,
                 )
 
-            # Compute scores for each turn, keyed by turn_id
+            # Audio-native rates one entry per tool call: normalize the judge's
+            # ``tool_call_id`` onto the shared ``turn_id`` key used by the loop below.
+            if context.is_audio_native:
+                for item in turn_evaluations:
+                    if isinstance(item, dict) and "turn_id" not in item and "tool_call_id" in item:
+                        item["turn_id"] = item["tool_call_id"]
```
gabegma: I find this a bit ugly - can't we just change it below so it uses `item.get("turn_id") or item.get("tool_call_id")`?

configs/prompts/judge.yaml
weiz9: You're right, I will either remove or combine them

configs/prompts/judge.yaml
weiz9: Will remove 7

tests/fixtures/metric_signatures.json
weiz9: gotcha, will do that from now on moving forward!

configs/prompts/judge.yaml
weiz9: correct, I will remove that

weiz9:
Hi @gabegma, thank you so much for the reviews! I have made the changes accordingly, please take a look when you get the chance and let me know how it looks, thank you!

gabegma (approved):
LGTM! :)

PR #187, opened 2026-07-15
Fix prompts for audiollm for misrepresenting_tool_result and information_loss

katstankiewicz:
- bug where judge was told that the output of audio llm was the users transcription when it is the actual llm output.
  - `information_loss` and `misrepresenting_tool_result` implied that the text was a transcript for audio llm so to not judge if mistakes are made, but this is inaccurate for audio llms since it is the actual output. If the output is incorrect assume it is a model issue
- Faithfulness - removed "The bar for disambiguation is higher than for a text-based system because the assistant knows it is working from audio and should anticipate mishearings." because it is ambiguous

### DIFF
--- a/src/eva/metrics/base.py
+++ b/src/eva/metrics/base.py
@@ -147,7 +147,27 @@ class MetricContext:
 
     @property
     def is_audio_native(self) -> bool:
+        """True when the assistant hears user audio directly (S2S or AudioLLMs).
+
+        Use for judge fragments about *audio perception* (e.g. disambiguation
+        expectations): both S2S and AUDIO_LLM take raw audio in, so both own their
+        mishearings. Do NOT use for fragments that reason about how assistant turns
+        were captured in the trace — see ``is_s2s``.
+        """
         return self.pipeline_type in (PipelineType.S2S, PipelineType.AUDIO_LLM)
+
+    @property
+    def is_s2s(self) -> bool:
+        """True only for speech-to-speech models not AudioLLMs.
+
+        AUDIO_LLM emits text before a separate TTS step, so its assistant turns are
+        the model's literal output — character-level slips there are real, not
+        transcription artifacts. Judge fragments that excuse token-level
+        discrepancies as TTS/STT artifacts (misrepresentation / information-loss
+        carve-outs, the assistant-turns disclaimer) must gate on this, not on
+        ``is_audio_native``.
+        """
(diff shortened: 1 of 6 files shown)


src/eva/metrics/base.py
```diff
@@ -147,8 +147,28 @@ def __init__(
...
+        """True only for speech-to-speech models not AudioLLMs.
+
+        AUDIO_LLM emits text before a separate TTS step, so its assistant turns are
+        the model's literal output — character-level slips there are real, not
+        transcription artifacts. Judge fragments that excuse token-level
+        discrepancies as TTS/STT artifacts (misrepresentation / information-loss
+        carve-outs, the assistant-turns disclaimer) must gate on this, not on
+        ``is_audio_native``.
+        """
+        return self.pipeline_type == PipelineType.S2S
```
fanny-riols, line 170: That won't match ElevenLabs agents that is marked as s2s, right?

katstankiewicz: No, we switch elevenlabs agents pipeline type to cascade earlier in the process so it is cascade when it reaches here

tests/unit/metrics/test_faithfulness.py
```diff
@@ -37,6 +37,32 @@ def test_get_prompt_variables_s2s(self):
...
+        variables = self.metric.get_prompt_variables(ctx, "transcript")
+        # Audio-native perception → S2S-style user/disambiguation framing.
+        assert "raw audio" in variables["disambiguation_context"]
+        assert "speech-to-speech" in variables["user_turns_disclaimer"]
+        # Trace provenance is text, not audio → no free pass on character-level slips.
+        assert variables["misrepresentation_pipeline_note"] == ""
+        assert "STT transcriptions of the agent" not in variables["assistant_turns_disclaimer"]
+        assert "intended text" in variables["assistant_turns_disclaimer"]
+
+    def test_disambiguation_context_drops_comparative_bar(self):
```
gabegma: This test is a bit weird to me - I don't think we need to test its not longer there.

gabegma (approved):
Excellent, thank you!! I wanted to do the change from `audio_native` to `s2s` for some time now for those 2 metrics for the assistant turns only, this is perfect 👌

tests/unit/metrics/test_faithfulness.py
katstankiewicz: yeah good point! i'll remove

PR #191, opened 2026-07-22
Add sub metrics for conversation correctly finished

gabegma:


src/eva/utils/conversation_correctly_finished/classifier.py
```diff
@@ -0,0 +1,201 @@
...
+2. **Classification** (pure logic, no I/O) — order the per-cause ``detect_*`` functions by priority
+   and return the single primary cause (or ``None`` when the record isn't a parent failure).
+3. **Sub-metric builders** — turn a ``Classification`` / final-turn text into ``MetricScore`` flags.
+"""
+
+import re
+from pathlib import Path
+from typing import Any
+
+from eva.metrics.processor import is_agent_timeout_on_user_turn
```
katstankiewicz: i think we should move `is_agent_timeout_on_user_turn` to utils instead of importing from eva.metrics since the metrics already import utils or compute `is_parent_failure` in the metric itself

src/eva/utils/conversation_correctly_finished/final_turn.py
```diff
@@ -0,0 +1,132 @@
...
+    Heuristics mirror the analysis app (EVABench `apps/analysis.py`) so the two tools agree, and are
+    **English-only** (the ack lead, spoken-digit words, and "X as in Y" cues are English phrasing;
+    single-letter/digit runs, caps codes, and NATO are largely language-neutral). Prosody/annotation
+    tags ([slow], [neutral], …) inflate word counts, so they're stripped first.
+
+    - ``short``: 1-2 words (< 3), counted **after stripping `[annotation]` tags**.
+    - ``acknowledgement``: leads with a confirmation/ack ("Yes, that is correct.", "Ok thanks",
+      "[neutral] Okay.", "No.") AND is at most ``_ACKNOWLEDGEMENT_MAX_WORDS`` words — a long sentence
+      that merely starts "yeah …" is not counted.
+    - ``spelled_entity``: letter/digit spell-out of an ID/code/name ("E M P eight nine …", NATO,
```
katstankiewicz, line 123: i think the NATO alphabet check is a bit loose, eg "hotel" or "november". Like i think "I am waiting in the hotel lobby" or "I need a flight that leaves November 6th." would get flagged incorrectly

src/eva/metrics/diagnostic/conversation_correctly_finished.py
```diff
@@ -30,12 +46,24 @@ async def compute(self, context: MetricContext) -> MetricScore:
...
+            # cross-record mean has an all-records denominator, matching faithfulness.
             if missed_turn:
                 human_reason = "conversation ended with inactivity_timeout and user was the last speaker"
-            elif reason == "inactivity_timeout":
-                human_reason = f"inactivity_timeout but last speaker was {speaker!r}"
+                signals = extract_conv_finish_signals(context)
+                classification = classify_conv_finish_failure(signals)
+                sub_metrics = build_conv_finish_sub_metrics(classification, self.name)
+                # Orthogonal input-characteristic flags (short / acknowledgement / spelled final turn).
+                sub_metrics.update(build_final_turn_flag_sub_metrics(signals.user_final_words, self.name))
```
katstankiewicz: this section should likely have it's own try-catch, otherwise if something fails in here, we lose the other calculated info like `details.reason` or `details.speaker`

src/eva/utils/conversation_correctly_finished/classifier.py
```diff
@@ -0,0 +1,201 @@
...
+_RE_RESP = re.compile(r"complete response: '(.*)'$")
+
+
+# --- extraction (I/O) -------------------------------------------------------
+
+
+def _extract_audit_log(audit: Path, s: ConvFinishSignals) -> None:
+    """Last conversation message role, from audit_log.json."""
+    if not audit.exists():
+        return
```
katstankiewicz: minor but `load_audit_log` will return none if the path doesn't exist so we don't need the extra check
```suggestion
```

src/eva/utils/conversation_correctly_finished/final_turn.py
gabegma: Great point

src/eva/metrics/diagnostic/conversation_correctly_finished.py
gabegma: Done!

gabegma:
@katstankiewicz thank you for the feedback! I should have addressed all of your comments.

katstankiewicz (approved):
thank you! lgtm!

PR #192, opened 2026-07-22
Make assistant nudge user if they are silent for > n seconds

katstankiewicz:
- Use `EVA_TURN_END_FALLBACK_TIME` to set the number of seconds to wait before nudging user
- The assistant will not nudge if user is speaking

todo:
- Small bug fix - if user is speaking or started speaking after the nudge is triggered the assistant doesn't say anything but that turn still gets counted as nudged
- Weird scenario where assistant transcript is not being captured in conversation_trace (the assistant asked if the user was still there after the vad didn't fire on the short turn)
Screenshot 2026-07-22 at 6 13 45 PM
from https://snow-core_llm-eva_results_analysis_plusplus.job.console.elementai.com/record_detail?output_dir=%2Fmnt%2Fvoice_agent%2Fks&record=T6.2&latest_only=false

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -198,6 +198,14 @@ EVA_MODEL_LIST='[
 #r 30,10000,10
 #v EVA_CONVERSATION_TIME_LIMIT_SECONDS=600
 
+#i Seconds of user silence after the assistant stops speaking before nudging it to retry the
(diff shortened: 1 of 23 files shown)


raghavm243512 (approved):
thanks for catching the bugs

PR #193, opened 2026-07-23
Refactor orchestrator runner

JosephMarinier:
* [Clean up (trivial)](https://github.com/ServiceNow/eva/pull/193/commits/96a345b7f83ec5c841a71aa1323e276a3a3ab86f)
* [Remove unused `finished_ids`](https://github.com/ServiceNow/eva/pull/193/commits/50d986761231e7795b1a22c2110eacd90e870f04)
* Bring `run()` (normal run) on par with `validate_existing()` (rerun)
  * [Sync `evaluation_summary.json` between `run()` and `validate_existing()`](https://github.com/ServiceNow/eva/pull/193/commits/ddc37e005a8d8c3f2c4aa406bbc13e188d6b0562) by removing fields that are not present in both (so I guess they were not important, but I'm happy to restore any if you want), and allowing non-ASCII characters (which was only the case in `validate_existing()`).
    * [Adapt test](https://github.com/ServiceNow/eva/pull/193/commits/0a24f74110f07394843c16e21e77c796db50e222)
  * [Archive the final failing attempts after the last `check_conversation_finished()`](https://github.com/ServiceNow/eva/pull/193/commits/b894a6f71d1094322c60e7ceb1d14c5f1c7134b3) in `run()` like it's done in `validate_existing()`. In practice, this `check_conversation_finished()` was not reachable because failed records always had an entry in `rerun_history`, but that seemed fragile.
  * [Simplify categorizing failures](https://github.com/ServiceNow/eva/pull/193/commits/dc10c2ed6ccbfee16c90463e7ccf057c42d4180f) in `run()` like it's done in `validate_existing()`.
  * [Archive at the beginning of the attempt loop](https://github.com/ServiceNow/eva/pull/193/commits/a0726341973269f4a5de70eb9b41a1c108794a8d) in `run()` like it's done in `validate_existing()`.
  * [Always await audio for incomplete conversations](https://github.com/ServiceNow/eva/pull/193/commits/208b8c860197b74bdf3228352ba026808012fa6f) 
    Remove the early-return for incomplete conversations in `_run_and_pipeline()`. It returned before `await audio_task`, leaving the audio task running un-awaited. The pipeline then proceeded without flushing the WAV and any exception went unobserved. This mattered most for `time_limit_exceeded` records, which are later skip-gate validated and can be accepted, so their audio must be written first.

    Classification is unchanged: an unfinished conversation still fails the `conversation_valid_end` gate in `validate_one()` and is bucketed `not_finished`.
    * [Adapt test](https://github.com/ServiceNow/eva/pull/193/commits/56e5bea5bd3180ca37b6ccadc69b43af7c8da405)
  * [Write `result.json` fallback when the worker skips it](https://github.com/ServiceNow/eva/pull/193/commits/ec9da6deec4a633a21ca0256769ac76b38c88f9e) 
    The worker writes `result.json` itself on the happy path, but its post-processing-exception path returns a result without writing, relying on the runner to do it. The worker code includes this comment:
    ```
    # Return a failed result so result.json still gets written by the
    # runner, rather than raising and leaving no result.json on disk.
    ```
    But, `run()` didn't actually write it, until now.
* [Share code between `run()` and `validate_existing()`](https://github.com/ServiceNow/eva/pull/193/commits/9865b4724722534821b819d5e1add81db85a5c21) 
    Extract `run()`'s attempt loop into `_run()` and reuse it in `validate_existing()`. This removes duplicated code, and `validate_existing()` gains a few features that only `run()` had:

    - per-record pipelining: metrics fire as soon as a conversation completes instead of waiting for all conversations to complete.
    - time-limit acceptance (added as "timeout" acceptance in b723a0fb, then renamed to "time-limit" in 1a427867) now also applies during reruns, not just the initial `run_validation()` pass.
    - a progress bar
* [Add run_id to progress bar description](https://github.com/ServiceNow/eva/pull/193/commits/82bf3bd0bc1d817ce9990417d625748af6554e05) so we can differentiate multiple runs launched at once.

### DIFF
--- a/src/eva/orchestrator/runner.py
+++ b/src/eva/orchestrator/runner.py
@@ -154,8 +154,7 @@ class BenchmarkRunner:
         # Apply record filtering (debug mode or specific record IDs)
         filtered_records = self._filter_records(records)
 
-        self.output_dir.mkdir(parents=True, exist_ok=True)
-        (self.output_dir / "records").mkdir(exist_ok=True)
+        (self.output_dir / "records").mkdir(parents=True, exist_ok=True)
 
         # Resolve exact models used (captures defaults from services.py + any alias labels)
         if self.config.model.pipeline_type == PipelineType.CASCADE:
@@ -207,10 +206,10 @@ class BenchmarkRunner:
 
         # Pre-create ValidationRunner for per-record pipelining. Its shared MetricsRunner
         # for validation metrics is lazily initialized on first use.
-        _all_unique_records = list({id(r): r for r in output_id_to_record.values()}.values())
+        all_unique_records = list({id(r): r for r in output_id_to_record.values()}.values())
         pipeline_validation_runner = ValidationRunner(
             run_dir=self.output_dir,
-            dataset=_all_unique_records,
+            dataset=all_unique_records,
             thresholds=self.config.validation_thresholds,
             metric_configs=self._metric_configs,
         )
@@ -221,7 +220,6 @@ class BenchmarkRunner:
         metrics_background_tasks: list[asyncio.Task] = []
 
         if self.config.metrics:
-            all_unique_records = list({id(r): r for r in output_id_to_record.values()}.values())
             try:
                 metrics_runner = MetricsRunner(
                     run_dir=self.output_dir,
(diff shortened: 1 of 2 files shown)


raghavm243512 (approved):
lgtm but we could just keep total attempts? we are still tracking the variable for it

src/eva/orchestrator/runner.py
```diff
@@ -538,9 +552,6 @@ async def _run_and_pipeline(
                         "validation_failed_count": validation_failed_count,
                         "records_with_llm_generic_error": len(llm_generic_error_record_ids),
                         "llm_generic_error_record_ids": llm_generic_error_record_ids,
-                        "success_rate": successful_count / total_tasks if total_tasks > 0 else 0.0,
-                        "failure_rate": failed_count / total_tasks if total_tasks > 0 else 0.0,
-                        "total_attempts": attempt_number,
```
JosephMarinier, line 543: @raghavm243512
> we could just keep total attempts? we are still tracking the variable for it

Yes, we still have `attempt_number`. On the initial run, that's pretty straightforward. However, on reruns, it's not as straightforward. Before this PR, `validate_existing()` was dropping the `total_attempts` field and instead adding this:
```py
"total_rerun_attempts": max(len(v) for v in rerun_history.values()) if rerun_history else 0,
```
which is almost the same `attempt_number`, except in one edge case: if no records needed a rerun at all (`pending_ids` empty from the start), the `for` loop was still setting `attempt_number = 1` before immediately breaking (runner.py:785-786), but `rerun_history` stayed empty, so `total_rerun_attempts` was `0` while `attempt_number` was `1`. Anyway, it was not really a "total", wasn't it?

Ideally, I guess `total_attempts` would increment from the previous run, but:
1. Simply adding `total_attempts` would be a naive "prior max + this run's max," and not a true per-record running total.
2. A true per-record running total would require saving per-record attempt counts from all previous runs, or listing/parsing many more files, which seems fragile.
1. I'm not sure this information is worth this effort. Do you have any use for it that I don't see?

What do you think?

raghavm243512: Yea I guess not a big deal, I'll just merge then

PR #195, opened 2026-07-27
Refactor/phase1 code path consolidation

raghavm243512:
Combing code that is separate and could instead be reused.

This will enable backends to share tool calling implementation and logging

Also updated the backend / role classes to match the newly added fallback. They are still skeletons like #190

End to end testing done locally, everything is identical (though recording order changed slightly, so some tool calls on non S2S runs will be added to audit log a couple milliseconds earlier)


katstankiewicz (approved):
lgtm

PR #196, opened 2026-07-29
Add preflight

JosephMarinier:
* [Fix strict LLM check](https://github.com/ServiceNow/eva/pull/196/commits/cf40f51f9655a296b35cff71a0aad106a85e8bd4) 
  See https://github.com/ServiceNow/eva/commit/6886ad60224a18ee5db6805d5c64b66feb35895d for more details.
* [Suppress `usage:` block in `eva --help`](https://github.com/ServiceNow/eva/pull/196/commits/d7df899cf02a9a25f0157e0e35cd39d22e849d2f) as it is completely redundant with the options below.
* [Migrate deprecated pipecat runner to worker](https://github.com/ServiceNow/eva/pull/196/commits/0e732b62e513e46adfa11e175b74025b26aad914)
* [Add preflight](https://github.com/ServiceNow/eva/pull/196/commits/5620ab0a005ed159c509b342bca9b61431677ac7) 
  Probe each configured model before a run starts and abort early on a credential/connectivity failure, instead of discovering it only after every record has exhausted its attempts.
  - eva/orchestrator/preflight.py: run_preflight probes the run's models by reusing the real client/factory paths (LiteLLMClient for the LLM, the pipecat STT/TTS service factories, the audio-LLM client). It is conservative, so only a definitive error fails a probe. S2S has no live probe yet (logged).
  - Wired into BenchmarkRunner._run() — the single point both a fresh run and a rerun funnel through — gated on "there's something to run" (pending records + attempts remaining), so it fires exactly once per invocation and only when conversations will actually run.
  - Added configs `--no-preflight` and `--preflight-timeout-seconds`.

### DIFF
--- a/src/eva/models/config.py
+++ b/src/eva/models/config.py
@@ -51,7 +51,7 @@ def _get_all_metrics() -> list[str]:
     return [m for m in get_global_registry().list_metrics() if m not in _VALIDATION_METRIC_NAMES]
 
 
-def _param_alias(params: dict[str, Any]) -> str:
+def get_model_alias_from_params(params: dict[str, Any]) -> str:
     """Return the display alias from a params dict."""
     return params.get("alias") or params["model"]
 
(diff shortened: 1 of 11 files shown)


src/eva/models/config.py
```diff
@@ -671,6 +669,12 @@ class ModelDeployment(DeploymentTypedDict):
         description="Logging level",
     )
     dry_run: bool = Field(False, description="Validate configuration without running")
+    preflight: bool = Field(
```
raghavm243512, line 672: should we have this as default false if it is described as "skip the check". Or just a phrasing error?

src/eva/models/config.py
```diff
@@ -671,6 +669,12 @@ class ModelDeployment(DeploymentTypedDict):
...
+        True,
+        description="Skip the pre-flight model check. By default, probe each configured model "
```
raghavm243512:
```suggestion
        description="Run the pre-flight model check. By default, probe each configured model "
```

raghavm243512 (approved):
looks good besides description

src/eva/models/config.py
JosephMarinier: Yeah, that's confusing. Because the default value of `preflight ` is `True`, and because we have `cli_implicit_flags="toggle"` in the `SettingsConfigDict`, Pydantic automatically defines the CLI argument as `--no-preflight`, which sets `preflight` to `False`, and effectively "skips" the preflight checks. This `description` isn't for the `preflight` field, but for the `--no-preflight` CLI argument. See `eva --help`:
```
--no-preflight        Skip the pre-flight model check. By default, probe each configured model (STT/LLM/TTS/audio-LLM) before the run starts, aborting early on credential/connectivity failures.
```
That said, you made me realize that I left a bunch on `pre-flight` with hyphens in comments, so I'll standardize that.
```suggestion
        description="Skip the preflight model check. By default, probe each configured model "
```
Does that make sense?

PR #199, opened 2026-07-30
add assistant gender

raghavm243512:
Add gender (in the form of a prompt addition) to assistant, so that it uses the right verbs and such in gendered languages matching whatever TTS it has. (no inferring from the TTS, this would have to be set when run)

### DIFF
--- a/src/eva/models/config.py
+++ b/src/eva/models/config.py
@@ -211,6 +211,20 @@ class ModelConfig(BaseModel):
         None,
         description="Forward parallel_tool_calls when tools are present; None leaves provider defaults.",
     )
+    assistant_gender: Literal["M", "F"] | None = Field(
+        None,
+        description=(
+            "Assistant speaking gender ('M' or 'F'), appended as a small system prompt "
+            "instruction so gendered languages (e.g. Hindi) produce grammatically consistent "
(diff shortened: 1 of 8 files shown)


src/eva/models/config.py
```diff
@@ -211,6 +211,20 @@ def _normalize_pre_tool_speech(cls, value: str) -> str:
         None,
         description="Forward parallel_tool_calls when tools are present; None leaves provider defaults.",
     )
+    assistant_gender: Literal["M", "F"] | None = Field(
+        None,
+        description=(
+            "Assistant speaking gender ('M' or 'F'), appended as a small system prompt "
+            "instruction so gendered languages (e.g. Hindi) produce grammatically consistent "
+            "speech. CASCADE/AUDIO_LLM only — S2S providers manage their own voice/gender. "
+            "None (default) adds no instruction, for gender-neutral languages like English."
```
JosephMarinier: Can we clarify this? One could read that `None` is only supported in gender-neutral languages like English, but it's always an option, right? Does that make sense?
```suggestion
            "None (default) adds no instruction, which is perfect for gender-neutral languages like English."
```

JosephMarinier (approved):
Cool! 👍

PR #200, opened 2026-07-30
move and rename

raghavm243512:
The "audio bridge" is not really an audio bridge at all and is really just a bunch of utils

I moved the log writers as well since those aren't exactly observers, but it's an ok place for them and they don't really need their own file


JosephMarinier (approved):
Good! 👍

PR #203, opened 2026-08-05
refactor proof of concept

raghavm243512:
This is the first of the main changes to how we organize servers and user simulators. 

The key here is rather than openai_realtime_user_simulator and openai_realtime_server.py, we just have user, assistant, and openai_realtime. This means every assistant backend will work as a user, including cascade architectures, ALMs, and all the modularity that comes with those. Future contributions will be far more useful and far easier to do.

All API interaction code is shared, meaning this design scheme will mean adding support for 1 API will work in any role, no more dedicated implementations for either role in the conversation. 

I started with openai realtime as it is the simplest, single part, no separate VAD and it was supported on both sides. 

This PR will not be merged as is, it features temporary test gates, and the config is too restrictive for what we actually need for the user simulator, requiring a moderate size redesign for the user to really work as we want it.

There will also be deletions for old files not done here. This is solely to review, build on top of, and hopefully figure out a nice way to merge it all without the temporary clutter in the code or configs. 

This has been tested with the new code path being used for both sides of the conversation and all results are identical

### General shape
As mentioned, we now have user and assistant objects and these each have an instance of "backend", which can be any backend EVA connects to
This entire construction process is done in the worker like before, which simply invokes a backend factory, and gives this backend to each role
The worker never interacts with any backend, each role abstracts the API details completely from the rest of the project and each Backend class completely abstracts the actual API from the Role classes


src/eva/backend/base.py
```diff
@@ -1,8 +1,9 @@
 """Abstract ``Backend`` contract: pure API/session exchange, no role knowledge.
 
-DESIGN ONLY (Step 1 of the refactor, see docs/refactor-step1.md). This module
-defines shapes, not behavior -- every method body is a stub. Nothing in
-``eva.assistant`` or ``eva.user_simulator`` depends on this yet.
+Step 1 of the refactor (see docs/refactor-step1.md). This is the live
+``Backend`` contract -- implemented by ``eva.backend.openai_realtime`` and
```
JosephMarinier:
```suggestion
This is the live ``Backend`` contract -- implemented by ``eva.backend.openai_realtime`` and
```

src/eva/assistant/pipeline/observers.py
```diff
@@ -390,6 +390,30 @@ def write(self, event_type: str, data: dict, timestamp_ms: int | None = None) ->
         except Exception as e:
             logger.error(f"Error writing framework log: {e}")
 
+    # Convenience wrappers over write(). Restored after commit 4bf4881 ("move and
+    # rename") dropped them while relocating FrameworkLogWriter from audio_bridge.py
+    # to this module, leaving openai_realtime_server / gemini_live_server calling
+    # methods that no longer existed.
```
JosephMarinier:
```suggestion
```

JosephMarinier (commented):
Not now, but once all the pieces are in place, let's try to clean up the comments about historical changes. We will no longer care if the code has been refactored, moved, renamed, restored, etc. A couple of examples I saw:

PR #212, opened 2026-08-18
update language code for kokoro phonemizer

katstankiewicz:
Main kokoro pipeline takes in "en-us" as a valid language code, but the phonemizer only accepts aliases like `a` for `en-us`

### DIFF
--- a/src/eva/assistant/pipeline/services.py
+++ b/src/eva/assistant/pipeline/services.py
@@ -485,19 +485,33 @@ def create_tts_service(
         logger.info(f"Using Kokoro TTS: {params['model']}")
         kokoro_tts = OpenAITTSService(
             api_key=api_key,
-            model=params["model"],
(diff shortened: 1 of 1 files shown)


raghavm243512:
the aliasing is done on the deployment side right? what phenomizer are you referring to

katstankiewicz:
> the aliasing is done on the deployment side right? what phenomizer are you referring to

the aliasing is working properly for the main pipeline, but later on we see this error:

| File "/app/api/src/services/text_processing/phonemizer.py", line 82, in create_phonemizer
| raise ValueError(f"Unsupported language code: {language}")
| ValueError: Unsupported language code: en-us

I don't know where `app/api/src/services/text_processing/phonemizer.py` is, but as is, it only accepts the single letter language code

raghavm243512 (approved):
not sure why this would have changed since we are using the same container, but this is for some odd reason what kokoro does

PR #213, opened 2026-08-18
multilingual prompt changes

raghavm243512:
small adjustments for tool use with native or latin script. Small judge and user prompt adjustment for this case too


configs/prompts/judge.yaml
```diff
@@ -342,6 +342,7 @@ judge:
...
+      - Trying the native-script rendering of a name after a Romanized/English spelling failed a lookup, or vice versa (e.g., retrying "山田" after "Yamada" did not match, or trying a Romanized transliteration after the native-script form failed) — this is the same disambiguation behavior as above, grounded in the user's spoken name, not a fabrication. **Exception:** if the user has already explicitly spelled the name out letter by letter (or corrected/confirmed a specific spelling) earlier in the conversation, the assistant must use that confirmed spelling exactly — deviating from it to try an alternate script or spelling at that point IS a fabrication, since the ambiguity was already resolved by the user.
```
tara-bogavelli: `**Exception:** if the user has already explicitly spelled the name out letter by letter (or corrected/confirmed a specific spelling) earlier in the conversation, the assistant must use that confirmed spelling exactly — deviating from it to try an alternate script or spelling at that point IS a fabrication, since the ambiguity was already resolved by the user.`

I'm concerned this might not work super well, are we sure we want it to flag every time it gets the spelling wrong as fabricating tool parameters? i think lots of times the user has technically already specified but the assistant still isn't sure and needs to try multiple spellings.

raghavm243512: The conversation trace that the judge sees should show the intended spelling in the trace, even if the transcription is wrong the trace will show that. If the trace explicitly says to do something (how to spell) and the agent doesn't follow, that is not listening to instructions from the user and not a transcription error or something similarly confounding

fanny-riols:
```suggestion
      - Trying the native-script rendering of a name after a Romanized/English spelling failed a lookup, or vice versa (e.g., trying "山田" after "Yamada" did not match, or trying a Romanized transliteration after the native-script form failed) — this is the same disambiguation behavior as above, grounded in the user's spoken name, not a fabrication. **Exception:** if the user has already explicitly spelled the name out letter by letter (or corrected/confirmed a specific spelling) earlier in the conversation, the assistant must use that confirmed spelling exactly — deviating from it to try an alternate script or spelling at that point IS a fabrication, since the ambiguity was already resolved by the user.
```

configs/prompts/judge.yaml
fanny-riols, line 351: We also had this note from Rachel for French, as something the faithfulness judge didn't catch:
> Misrepresenting tool results – one time the agent said Côté (pronunciating the `e` with the accent) even though the tool said the user’s last name was Cote.

[Sample](https://snow-core_llm-eva_results_analysis_plusplus.job.console.elementai.com/record_detail?output_dir=%2Fmnt%2Fvoice_agent%2Fmultilingual&record=52&run=fr-CA_nova-3_gpt-5.4_sonic-3.5%20%C2%B7%20itsm%20%C2%B7%20FR-CA%20%C2%B7%20Clean%20%282026-06-02_22-47-52.765659%29)

Image

Can you add something to fix that please?

tests/fixtures/metric_signatures.json
fanny-riols, line 1: I think we should also bump the simulator. Minor bump.

configs/prompts/judge.yaml
fanny-riols: Won't do for now, as discussed in standup.

configs/agents/itsm_agent.yaml
```diff
@@ -76,6 +76,8 @@ instructions: |
 
   Before making any change to a record, retrieve and review the current state of that record with the caller.
 
+  **Script ambiguity:** A caller-provided name (e.g., an application/license product name, or a manager's or new hire's name mentioned in conversation) may need to be queried in either its native script or a Romanized/English spelling. If a lookup or catalog match fails on the first attempt, try the alternate script or spelling before asking the caller to repeat themselves or treating it as an error. Once the caller has explicitly spelled a value out letter by letter, use that confirmed spelling exactly — do not substitute an alternate spelling at that point.
```
tara-bogavelli: `A caller-provided name (e.g., an application/license product name`
so the application license product names can also be expected in multiple scripts?

configs/prompts/simulation.yaml
```diff
@@ -299,8 +299,8 @@ user_simulator:
     - Spaces in codes → spoken with pauses between characters
     ## Backup Behavior
     If the agent mishears you and gets any information wrong, repeat it even slower and spell it out using a phonetic alphabet appropriate for your language, one character at a time. ONLY do this if the agent mishears you the first time.
-    For languages that use the Latin alphabet, use the NATO phonetic alphabet (e.g. "C as in Charlie, X as in X-ray").
-    For languages that use non-Latin scripts, spell out characters using their standard phonetic names in your language.
+    For English, use the NATO phonetic alphabet (e.g. "C as in Charlie, X as in X-ray").
+    For other languages, spell out characters using their standard phonetic names in your language.
```
tara-bogavelli: i think the linguists said this depends on language though, like in some languages they still use English letter spelling but in others they wouldn't.

fanny-riols: Yeah but I understand "their standard phonetic names" as refering to whatever people use in that language, what's standard to them. No?

But, I also don't find this 2nd sentence super clear. Suggestion:
```suggestion
    For other languages, spell out characters using the standard phonetic alphabet or character names used in that language.
```

configs/agents/itsm_agent.yaml
raghavm243512: yes, but they have translations in native script + English names as valid lookup candidates. Sometimes a name won't have a direct translation, sometimes it will. For example: "headquarters" would be translated, "IntelliJ" would not. This actually is something I will ask linguists about for 'transliteration', I may add transliterations to the un-translated data.

tara-bogavelli (approved):
LGTM thanks!

PR #223, opened 2026-09-02
Support overriding metric configs

JosephMarinier:
Support overriding metric configs via CLI arguments or environment variables.

Examples:

`EVA_METRIC_CONFIGS='{"faithfulness": {"judge_model": "gpt-5.6-terra", "judge_params": {"reasoning_effort": "none"}}}'`
`EVA_METRIC_CONFIGS__FAITHFULNESS__JUDGE_MODEL=gpt-5.6-terra`
`EVA_METRIC_CONFIGS__FAITHFULNESS__JUDGE_PARAMS__REASONING_EFFORT=none`

### DIFF
--- a/src/eva/models/config.py
+++ b/src/eva/models/config.py
@@ -13,6 +13,7 @@ and explicit kwargs.  Scripts opt in to ``.env`` and/or CLI via
 """
 
 import copy
+import json
 import logging
 import os
 from collections.abc import Iterator
@@ -609,6 +610,18 @@ class RunConfig(BaseSettings):
     metrics: list[str] | None = Field(
         default_factory=_get_all_metrics,
         description="Metrics to run. Skip all metrics with `EVA_METRICS=` or `--metrics=`.",
+    )
+
+    metric_configs: dict[str, dict[str, Any]] = Field(
+        {},
+        description=(
+            "Per-metric configuration overrides (JSON), keyed by metric name. For judge metrics "
+            "(TextJudgeMetric/AudioJudgeMetric subclasses), 'judge_model' overrides the model "
+            "and 'judge_params' merges into the LLM call params (e.g. reasoning_effort, temperature). "
+            f"Example: EVA_METRIC_CONFIGS='{json.dumps({'faithfulness': {'judge_model': 'gpt-5.6-terra', 'judge_params': {'reasoning_effort': 'none'}}})}' "
+            "or EVA_METRIC_CONFIGS__FAITHFULNESS__JUDGE_MODEL=gpt-5.6-terra "
+            "or EVA_METRIC_CONFIGS__FAITHFULNESS__JUDGE_PARAMS__REASONING_EFFORT=none."
+        ),
     )
 
     # Aggregate-only mode
--- a/src/eva/run_benchmark.py
+++ b/src/eva/run_benchmark.py
@@ -55,6 +55,8 @@ async def run_benchmark(config: RunConfig) -> int:
         runner.config.preflight_timeout_seconds = config.preflight_timeout_seconds
         if config.metrics is not None:
             runner.config.metrics = config.metrics
+        if config.metric_configs:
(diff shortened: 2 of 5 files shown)


src/eva/run_benchmark.py
```diff
@@ -55,6 +55,8 @@ async def run_benchmark(config: RunConfig) -> int:
         runner.config.preflight_timeout_seconds = config.preflight_timeout_seconds
         if config.metrics is not None:
             runner.config.metrics = config.metrics
+        if config.metric_configs:
+            runner.config.metric_configs = config.metric_configs
```
JosephMarinier: We can debate this. Should it be possible to override metric configs in a rerun? It was useful to me for testing, but it might be too dangerous to accidentally override metric configs.

JosephMarinier: I think that's a debate we could have about multiple config fields. Instead, we could consider adding an option to explicitly enable config overrides (or disable reading the config saved on disk), something like:
`eva --run-id ... --ignore-previous-config`

katstankiewicz: i like the --ignore-previous-config idea. For most config values right now I don't think we can override them on re-run, so I wonder if we should enable overriding all of them at the same time and introduce `--ignore-previous-config` at the same time

JosephMarinier: I'll create a separate PR after this one to introduce `--ignore-previous-config`. For now, I'll remove this override. Does that make sense? And do the changes from this PR make sense?
```suggestion
```

JosephMarinier: - Here is a separate PR to #229

gabegma (approved):
LGTM! Should we document somewhere?

src/eva/models/config.py
```diff
@@ -611,6 +612,18 @@ class ModelDeployment(DeploymentTypedDict):
...
+    metric_configs: dict[str, dict[str, Any]] = Field(
+        {},
+        description=(
+            "Per-metric configuration overrides (JSON), keyed by metric name. For judge metrics "
+            "(TextJudgeMetric/AudioJudgeMetric subclasses), 'judge_model' overrides the model "
+            "and 'judge_params' merges into the LLM call params (e.g. reasoning_effort, temperature). "
+            f"Example: EVA_METRIC_CONFIGS='{json.dumps({'faithfulness': {'judge_model': 'gpt-5.6-terra', 'judge_params': {'reasoning_effort': 'none'}}})}' "
+            "or EVA_METRIC_CONFIGS__FAITHFULNESS__JUDGE_MODEL=gpt-5.6-terra "
+            "or EVA_METRIC_CONFIGS__FAITHFULNESS__JUDGE_PARAMS__REASONING_EFFORT=none."
+        ),
```
JosephMarinier: Thank you @gabegma! 🙂 

> LGTM! Should we document somewhere?

This `description` here can be seen in `eva --help`. Is that good enough?

I know the main env vars are also documented in `.env.example` as a "quick start" guide. Some more niche/experimental config fields, like `run_id`, `num_trials`, `aggregate_only`, `ignore_previous_config`, `force_rerun_metrics`, `tool_module_path`, and `validation_thresholds`, are not documented in `.env.example`. I assumed that was intentional to avoid overwhelming `.env.example` and duplicating documentation for all fields, which would increase the risk of drifting out of sync, but maybe it was just an oversight.

I pushed a commit to [Add field descriptions to `.env.example`](https://github.com/ServiceNow/eva/pull/223/commits/1903b0ab552673b5661c02305b378a5dd6865448). Does that make sense to you? I can easily revert if you had something else in mind.

gabegma: Ok I'm ok with eva --help! I thought we might have documentation elsewhere. I'm not particularly a fan of the documentation in `.env.example` personally

JosephMarinier: Agreed. I'll revert my commit to `.env.example` then. Thank you!

PR #226, opened 2026-09-15
Feat/gemini eval

raghavm243512:
Bring over changes from google for new Gemini Live model. Original branch was conflicting and pretty old, so this has all the changes integrated

Some changes related to tool scheduling for new gemini live models, a couple additional diagnostic metrics, and a white noise addition that I disabled since we don't have global testing with this change (but is easy to turn on). 

I also updated the Dockerfile and this small reorder should have better reuse of built content, so smaller updates when we make changes


README.md
```diff
@@ -97,6 +97,7 @@ EVA_DOMAIN=airline
 EVA_MAX_CONCURRENT_CONVERSATIONS=5
 EVA_DEBUG=false                       # Run only 1 record for testing when enabled
 EVA_RECORD_IDS=1.2.1,1.2.2            # Run specific records only (remove to run all records)
+EVA_EXCLUDE_RECORD_IDS=6.1.1          # Skip specific records (applied after EVA_RECORD_IDS)
```
fanny-riols:
```suggestion
```

Dockerfile
```diff
@@ -101,6 +91,16 @@ ENV PYTHONUNBUFFERED=1
...
 
+# Git provenance baked in at build time
+ARG GIT_COMMIT_SHA
+ARG GIT_BRANCH
+ARG GIT_DIRTY
+ARG GIT_DIFF_HASH
+ENV GIT_COMMIT_SHA=${GIT_COMMIT_SHA}
+ENV GIT_BRANCH=${GIT_BRANCH}
+ENV GIT_DIRTY=${GIT_DIRTY}
+ENV GIT_DIFF_HASH=${GIT_DIFF_HASH}
```
fanny-riols, line 102: What does it change to move it there?

raghavm243512: when these values change, the build stages get invalidated and are rebuilt and reuploaded. If they are earlier in the build (old place) they invalidate subsequent things and we waste time uploading and rebuilding. In general, docker files should start with heavy things first, and light things later (especially if the light things change a lot)

src/eva/assistant/gemini_live_server.py
```diff
@@ -188,11 +207,68 @@ def __init__(
...
+        #   "thinking_level": str   ("MINIMAL", "LOW", "MEDIUM", "HIGH")
+        # Example: {"thinking_budget": 1024, "include_thoughts": false}
+        # If not set, defaults to ThinkingConfig() (model-dependent defaults).
+        thinking_raw = s2s_params.get("thinking_config", {})
+        if isinstance(thinking_raw, dict) and thinking_raw:
+            tc_kwargs: dict[str, Any] = {}
+            if "thinking_budget" in thinking_raw:
+                tc_kwargs["thinking_budget"] = int(thinking_raw["thinking_budget"])
+            if "include_thoughts" in thinking_raw:
+                tc_kwargs["include_thoughts"] = bool(thinking_raw["include_thoughts"])
```
katstankiewicz: should we include_thoughts by default if thinking_level is set? unless include_thoughts is set to false?

src/eva/assistant/gemini_live_server.py
```diff
@@ -188,11 +207,68 @@ def __init__(
...
+            if "thinking_level" in thinking_raw:
+                tc_kwargs["thinking_level"] = thinking_raw["thinking_level"]
+            self._thinking_config = types.ThinkingConfig(**tc_kwargs)
+            logger.info(f"Thinking config: {tc_kwargs}")
+        else:
+            self._thinking_config = types.ThinkingConfig()
```
katstankiewicz: i don't think include_thoughts or thinking_budget are supported by live models so they can be removed

src/eva/assistant/gemini_live_server.py
```diff
@@ -138,7 +157,7 @@ def _agent_tools_to_gemini(agent: AgentConfig) -> list[types.Tool] | None:
                 name=tool.function_name,
                 description=f"{tool.name}: {tool.description}",
                 parameters=params_schema,
-                behavior=types.Behavior.BLOCKING,
+                # behavior=types.Behavior.BLOCKING,
```
katstankiewicz: the blocking behaviour is supported by gemini 3.8 live but not 3.8 live extended thinking, so can we either enable for the or remove this line if it is not needed anymore

src/eva/metrics/diagnostic/time_to_completion.py
```diff
@@ -0,0 +1,77 @@
...
+"""
+
+from eva.metrics.base import CodeMetric, MetricContext
+from eva.metrics.diagnostic.task_completion_utils import is_task_completed
+from eva.metrics.registry import register_metric
+from eva.models.results import MetricScore
+
+
+@register_metric
+class TimeToCompletionMetric(CodeMetric):
```
katstankiewicz: i think time_to_completion and conversation_completed_on_time are quite similar. what do you think about having `conversation_completed_on_time` and `timed_out` as submetrics for this one and remove the metric `conversation_completed_on_time`?

katstankiewicz: Also, i'm wondering what this is measuring exactly since the audio is often much shorter that this time to completion. I think it includes some of the set up/tear down time which we probably dont want included in this metric

README.md
raghavm243512: Applied — removed the EVA_EXCLUDE_RECORD_IDS line, it wasn't actually implemented in code.

src/eva/assistant/gemini_live_server.py
raghavm243512: Removed thinking_budget support (kept include_thoughts + thinking_level). Also wired up capturing thoughts: when include_thoughts is set, parts with part.thought=True are now logged via a new FrameworkLogWriter.thought() call (framework_logs.jsonl, type="thought") instead of being silently dropped.

src/eva/assistant/gemini_live_server.py
raghavm243512: Checked — Behavior.BLOCKING/NON_BLOCKING is explicitly documented as unsupported in Vertex AI in the google-genai SDK, and this server always creates its client with vertexai=True (see _create_genai_client, Vertex-only Live-preview models require it). Since Vertex is our only path here, I removed the line rather than conditionally enabling it, and left a comment explaining why.

src/eva/metrics/diagnostic/time_to_completion.py
raghavm243512: Agreed, merged them: conversation_completed_on_time is now removed as a standalone metric, and time_to_completion's details now include conversation_completed_on_time and timed_out as submetrics (regenerated tests/fixtures/metric_signatures.json accordingly).

raghavm243512: Confirmed — duration_seconds is wall-clock started_at to ended_at from ConversationResult (orchestrator/worker.py), which does include connection setup/teardown, not just spoken audio. Updated the docstring to call this out explicitly. Left the underlying measurement as-is since narrowing it to pure audio duration would need a separate timing source and felt out of scope for this PR — happy to file a follow-up if that's wanted.

katstankiewicz (approved):
thank you for addressing all of the comments!

PR #229, opened 2026-09-18
Add option to ignore previous config on rerun

JosephMarinier:
From `eva --help`:
> --ignore-previous-config
> Ignore the config.json saved by the previous run (assuming --run-id points to an existing run). Use with caution. Without this flag, only a few fields (rerun, preflight, and metrics) can be overridden on rerun; the rest come from the saved config.json. (default: False)


src/eva/run_benchmark.py
```diff
@@ -38,23 +38,26 @@ async def run_benchmark(config: RunConfig) -> int:
...
     if existing_run:
         # ── Existing run: validate, rerun, or metrics-only ──
-        try:
-            runner = BenchmarkRunner.from_existing_run(resolved_dir)
-        except FileNotFoundError as e:
-            logger.error(str(e))
-            return 1
+        if config.ignore_previous_config:
+            runner = BenchmarkRunner(config)
+        else:
```
JosephMarinier, line 43: That's the only new bit; the rest of the changes in this file are just indentation.

raghavm243512 (approved):

PR #230, opened 2026-09-18
Add an optional `run_id_suffix` config

JosephMarinier:
Add an optional `run_id_suffix` config to distinguish different runs. Until now, we have been hacking a suffix into the run ID by overriding the alias of the last pipeline part (TTS or S2S), which can be error-prone.


raghavm243512 (approved):
looks good, maybe a nice name for this could be "experiment_tag"?

Super minor ofc, but I like when things explain their purpose in their name

src/eva/models/config.py
```diff
@@ -553,6 +553,7 @@ class ModelDeployment(DeploymentTypedDict):
         "timestamp and model name(s)",  # Overwritten by _set_default_run_id()
         description="Run identifier, auto-generated if not provided",
     )
+    run_id_suffix: str = Field("", description="Optional string appended to auto-generated run identifier")
```
JosephMarinier: Thanks, @raghavm243512, for the feedback!
> I like when things explain their purpose in their name

Agree! 👍 It's worth reflecting on the name. Thank you!

I'm hesitant to introduce the notion of "experiment". We usually refer to a "run", hence the `run_id`. To me, `experiment_tag` lacks conveying WHAT it does or WHERE you'll see it. Does it only go in the `config.json`? I appreciate that `run_id_suffix` clearly conveys that it's going to be appended to the `run_id`.

But I agree with you that "suffix" doesn't help with the purpose or WHY using it. And it's not really important if it goes in the suffix, the prefix, or somewhere in the middle. 🤔 

Here are a few ideas, with some new combinations:
- `run_id_suffix`
- `experiment_tag`
- `run_id_tag` or `rud_tag`
- `run_id_label` or `run_label`
- `run_id_note` or `run_note`
- `run_id_extra` or `run_extra`

It's hard to find a short name that conveys the what, where, and why. I guess the missing ones simply need to be clear in the `eva --help` documentation.

What do you think?

raghavm243512: I like run_label or run_tag

JosephMarinier: Thanks! See [Rename `run_id_suffix` to `run_tag`](https://github.com/ServiceNow/eva/pull/230/commits/9b572ecdaa562101f280d1478c2dd1099783892d).

